Issue #6, opened 2017-01-30
kotct/user-update-config: Always ask for a username


rye:
I'm thinking `M-x kotct/user-update-config` should behave a lot like `C-x C-z` in that it should ask the user for a username to fetch rather than just doing the current one.  I could be wrong about this one, however.

cg505:
I don't get it. Why?  What use case is the

cg505:
ugh sorry, misclicked

cg505:
What use case is this capturing?

rye:
For like when you want to update someone else's thing

cg505:
We could prompt on prefix arg.

rye:
Yeah, that's what I was going for.

PR #8, opened 2017-01-31
Prompt for Username On User Config Update

rye:
This allows us to update other users' configurations when they are not selected.

There are probably some things missing with this, so I welcome feedback.


cg505:
This doesn't match what we talked about in #6. Specifically, this always prompts for a username on update. That said, I think it's okay since the default is the current username. So 👍 .

cg505:
Geez this is actually super clean. Merging.

PR #12, opened 2017-02-01
Remove 'pry' as a dependency

### LINKED ISSUE #11: install fails without pry
if the install script is going to use pry, it ought to install it

rye:
This PR resolves #11.


cg505:
lol we didn't even need it?

rye:
nope

PR #14, opened 2017-02-01
Turn on hl-todo-mode with prog-mode-hook

rye:
Forgot to do this.


cg505 (changes requested):
You need to add `hl-todo` as a dependency.

rye:
FAIL. Will fix.

cg505 (approved):

PR #15, opened 2017-02-01
Add highlight-symbol

cg505:
This is a nice way to navigate through code that I use a lot, specifically `M-p` and `M-n`.


.emacs.d/lisp/code/code-hub.el
```diff
@@ -3,6 +3,7 @@
 (kotct/hub "code"
            (magit-c
             indentation
+            code-navigation
```
rye, line 6: I think this file could best be renamed to just `navigation` (I will fix this shortly in a commit that I am about to push.)

.emacs.d/lisp/code/code-navigation.el
```diff
@@ -0,0 +1,12 @@
+;;; M-n: go the next occurence of the symbol under point
+;;; M-p: go the previous occurence of the symbol under point
+;;;   note: these keybindings are enabled as a part of `highlight-symbol-nav-mode'
+;;; M-': replace all following occurences of the symbol under point
+
+;; enable highlight-symbol mode for all programming modes
+(add-hook 'prog-mode-hook (lambda ()
+                            (highlight-symbol-mode +1)
```
rye, line 8: Is there any particular reason to use `+1` vs `t`?

.emacs.d/lisp/code/code-navigation.el
```diff
@@ -0,0 +1,12 @@
...
+                            (highlight-symbol-nav-mode +1)
```
rye, line 9: Is there any particular reason to use `+1` vs `t`?

rye (changes requested):
Just a minor gripe with naming which I will take care of.

rye (approved):

.emacs.d/lisp/code/code-navigation.el
cg505: Yes. Modes enable on positive arguments and disable on non-positive arguments. To make this clear, it's a common practice to enable using +1 and disable using -1.

cg505 (commented):
The reason I didn't do this was I wanted to leave open the feature name `navigation` in case we need it for a non-code version. That said, I'm find with just calling it `navigation` if you don't think it'll be a problem.

.emacs.d/lisp/code/code-navigation.el
rye: Ah, thanks for the clarification.

rye:
Oh, I see your point.  I will revert that.  I think it's better to be clear.

rye:
Looks good now.

rye:
Merging.

PR #16, opened 2017-02-01
Add automated tab system

cg505:
Instead of `(setf var global-tab-width)`, do `(kotct/setf-tab var)`.

Instead of `(setq-default var global-tab-width)`, do `(kotct/setq-default-tab var)`.

Update the tab width with `kotct/set-tab-width`.

Is this kinda what you had in mind, @rye?


rye (approved):
This looks good.  We might refine it later, but I like this as-is. 👍

cg505:
Okay.
There are some bugs, so I'm not merging yet.

cg505:
Fixed bugs. How does it look now?

rye (approved):
Looks good!

PR #18, opened 2017-02-01
EditorConfig

rye:
This probably sucks and isn't exactly what we want, especially with the core warning, but it does implement basic support for `editorconfig-mode`.  I have tested it and it does effectively change indentation style.

### DIFF
--- a/.emacs.d/lisp/code/editorconfig-c.el
+++ b/.emacs.d/lisp/code/editorconfig-c.el
@@ -1,1 +1,4 @@
+;; Enable EditorConfig mode
+(editorconfig-mode +1)
+
 (provide 'editorconfig-c)
--- a/.emacs.d/lisp/code/editorconfig-c.el
+++ b/.emacs.d/lisp/code/editorconfig-c.el
@@ -9,4 +9,19 @@
 
 (add-hook 'editorconfig-custom-hooks 'kotct/editorconfig--disable-smart-tabs-if-indent-tabs-mode-disabled)
 
+(defvar kotct/editorconfig--suppress-core-not-found-warning
+  nil
+  "If non-nil, warning on startup is suppressed.")
+
+(defun kotct/editorconfig-check-for-core (&optional suppress)
+  "Checks if EditorConfig C core is installed.
+
+Debug warning is suppressed if SUPPRESS is non-nil."
+
+  (if (not suppress)
+      (if (not (executable-find editorconfig-exec-path))
+          (lwarn '(editorconfig core) :debug "EditorConfig executable not found."))))
+
+(kotct/editorconfig-check-for-core kotct/editorconfig--suppress-core-not-found-warning)
+
 (provide 'editorconfig-c)
--- a/.emacs.d/lisp/code/editorconfig-c.el
+++ b/.emacs.d/lisp/code/editorconfig-c.el
@@ -1,4 +1,12 @@
 ;; Enable EditorConfig mode
 (editorconfig-mode +1)
 
+(defun kotct/editorconfig--disable-smart-tabs-if-indent-tabs-mode-disabled (&optional hash)
+  "Disables `smart-tabs-mode' if `indent-tabs-mode' is non-truthy."
+  (if (not indent-tabs-mode)
+      (progn
+        (smart-tabs-mode -1))))
+
+(add-hook 'editorconfig-custom-hooks 'kotct/editorconfig--disable-smart-tabs-if-indent-tabs-mode-disabled)
+
 (provide 'editorconfig-c)
--- a/.emacs.d/lisp/code/editorconfig-c.el
+++ b/.emacs.d/lisp/code/editorconfig-c.el
@@ -1,8 +1,11 @@
 ;; Enable EditorConfig mode
 (editorconfig-mode +1)
 
-(defun kotct/editorconfig--disable-smart-tabs-if-indent-tabs-mode-disabled (&optional hash)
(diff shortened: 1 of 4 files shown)


.emacs.d/lisp/code/editorconfig-c.el
```diff
@@ -0,0 +1,27 @@
+;; Enable EditorConfig mode
+(editorconfig-mode +1)
+
+(defun kotct/editorconfig--disable-smart-tabs-if-indent-tabs-mode-disabled (&optional hash)
```
cg505: Please can we shorten some of the names? This is a little excessive.That's what the docstring's for.

cg505: Also what's the point of HASH?

.emacs.d/lisp/code/editorconfig-c.el
```diff
@@ -0,0 +1,27 @@
...
+      (progn
+        (smart-tabs-mode -1))))
+
+(add-hook 'editorconfig-custom-hooks 'kotct/editorconfig--disable-smart-tabs-if-indent-tabs-mode-disabled)
+
+(defvar kotct/editorconfig--suppress-core-not-found-warning
+  nil
+  "If non-nil, warning on startup is suppressed.")
+
+(defun kotct/editorconfig-check-for-core (&optional suppress)
```
cg505: Let's choose either `--` or `-` and stick to it. Personally I'm in favor of `-`.

.emacs.d/lisp/code/editorconfig-c.el
```diff
@@ -0,0 +1,27 @@
...
+  "Checks if EditorConfig C core is installed.
+
+Debug warning is suppressed if SUPPRESS is non-nil."
+
+  (if (not suppress)
+      (if (not (executable-find editorconfig-exec-path))
```
cg505, line 25: `editorconfig-exec-path` sounds like a path, not an executable name. If this is the case, we should be using something that checks if the file exists, not `executable-find`.

cg505 (commented):
I'm not at machine right now, so I haven't really tested this yet. I'll play around with it a bit this afternoon.

.emacs.d/lisp/code/editorconfig-c.el
rye: Do you have a suggestion for a better name?  `hash` is a hash passed by EditorConfig when it runs the hooks. It contains the loaded properties from the EditorConfig file.

.emacs.d/lisp/code/editorconfig-c.el
rye: No, it's an executable name. (i.e. `editorconfig`, no pathing)

.emacs.d/lisp/code/editorconfig-c.el
rye: Okay.  `-` works.  I generally use `--` for situations when I don't want people calling the functions, so really everything here should be `--`.

.emacs.d/lisp/code/editorconfig-c.el
cg505: It'd go with something like `kotct/editorconfig-maybe-disable-smart-tabs`.

cg505: Also the `hash` thing sounds fine. Maybe mention that it's only there for compatibility and is ignored in the docstring.

.emacs.d/lisp/code/editorconfig-c.el
cg505: I see. We haven't been using that convention in dot, but we could lol.

.emacs.d/lisp/code/editorconfig-c.el
cg505: Yikes. Bad name. Is that the editorconfig package's fault then?

.emacs.d/lisp/code/editorconfig-c.el
rye: We should.  Anything that we don't want users to use (i.e. stuff that isn't `(interactive)`-ified) should be with double dashes, imo.

cg505: Isn't the fact that it's not interactive enough?

rye: Aye, but we don't want the functions mixed in with the ones that are interactive when you're using something like an autocompleter.

.emacs.d/lisp/code/editorconfig-c.el
rye: Yes, but it makes sense.  This is platform-agnostic, and uses Emacs' builtin `which`-like mechanisms to hunt down the executable.  They should just use `editorconfig-executable`.

cg505: Yep. Looks good.

.emacs.d/lisp/code/editorconfig-c.el
cg505: Okay.
I'm open to this if you want to go through and change everything on a branch.

cg505: For now let's stick to `-`?

rye: I suppose.

rye:
@cg505 I'm going to go ahead and merge this.  As with most PR's, this is low-impact and adds functionality so it doesn't matter too much if it's broken.

PR #20, opened 2017-02-01
Set the standard-indent variable

rye:
This is used in a lot of places to specify standard indentation.  Web-mode, for one, uses it to determine the standard indentation.


cg505:
looks fine

PR #22, opened 2017-02-04
Kill loaddefs Buffer on Startup

### LINKED ISSUE #17: ~/.emacs.d/lisp/kotct-loaddefs.el is loaded on startup
This is annoying.  We shouldn't have this file be loaded into a buffer on startup.

rye:
This is a fairly trivial change, to resolve #17.  We just try and find the loaddefs buffer, and if we find it, we close it.

As opposed to just killing by filename, this throws in error handling so we don't get stupid warning panes.

### DIFF
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -10,7 +10,6 @@
 
 (setf generated-autoload-file "~/.emacs.d/lisp/kotct-loaddefs.el")
 
-
 ;;; hub initialization
 (defvar kotct/hub-list
   '("package"
@@ -52,6 +51,7 @@ If AUTOLOADS is non-nil, update the autoloads for that directory."
 
             (mapc #'require ,feature-var))))
 
+
 ;; add hub directories to load path
 (let ((default-directory "~/.emacs.d/lisp/"))
   (add-to-list 'load-path default-directory)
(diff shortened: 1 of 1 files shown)


.emacs.d/init.el
```diff
@@ -79,13 +79,21 @@ If AUTOLOADS is non-nil, update the autoloads for that directory."
   ;; start the process in *config-compilation* buffer
   (apply #'start-process args))
 
-
+;; Warn the user about shadowed files on the load path.  This usually
+;; happens when one keeps .elc files from removed .el files.
 (if (list-load-path-shadows)
     (message "There are shadowed files on your load path.
-This could indiciate an issue with your emacs installation.
+This could indicate an issue with your emacs installation.
```
cg505, line 83: whoops ty

cg505 (commented):
The actual code looks fine. Why the whitespace changes? The idea was that `;;;` indicates a new section, and each such comment is preceded by two blank lines.

rye:
Haha that wasn't in the spec so I didn't know. We can always go back to a cooler thing if you wanna

cg505:
I'd like to keep it following that unless you have a reason to change it. Also, it's not really significant enough to be a spec-level thing. It's more just something that came about as I was writing to keep it a bit more organized.

cg505 (approved):
lol ty

PR #23, opened 2017-02-04
Add a configuration for Darwin systems

rye:
This will help our various macOS Homebrew users (i.e. myself, possibly @samontea) deal with macOS' stupid PATH-related stuff and stupid modifier mappings.

Please note that there is very little way other than to directly parse the user's shell configuration ourself to get the PATH right, and on some shell configurations, this doesn't run very quickly.  This only affects Darwin users, however, and only really results in 0.5 more seconds in most common situations.


.emacs.d/init.el
```diff
@@ -18,7 +18,8 @@
     "visual"
     "behavior"
     "code"
-    "user")
+    "user"
+    "system")
```
cg505: `user` should always be the last hub.

.emacs.d/lisp/system/darwin.el
```diff
@@ -0,0 +1,44 @@
...
+  "This function swaps the Command and Option modifiers on MacOS.
+
+This is intended to make the keyboard layout more friendly for
+those used to a PC-like keyboard layout.  This makes C- and M-
+more spread apart and makes the keyboard more ergonomic.
+
+Because it uses the `mac-command-modifier' and
+`mac-option-modifier' builtins, it does not affect other systems
+built without compatibility for these two variables."
+  (progn
```
cg505: Can we `defvar` some variable that controls whether or not we do this or not?
Then call these commands after init (on a hook) to make sure that gets used?

.emacs.d/lisp/system/darwin.el
```diff
@@ -0,0 +1,44 @@
...
+In most sane environments, (i.e. Linux) the `exec-path' variable
+is suitably close to or identical to the 'PATH' used in a
+terminal.  On MacOS/Darwin environments, however, `exec-path' is
+only the global value for the variable 'PATH'.
+
+As such, access to things like Homebrew-installed packages is
+restricted.  This function calls the shell and asks it to print
+out the 'PATH' variable after it loads the `.profile'-related
+stuff.  This also respects the user's shell decision."
+
```
cg505: I don't like this newline.

.emacs.d/lisp/system/darwin.el
```diff
@@ -0,0 +1,44 @@
...
+  ;; Only run this once.  After you perform this step, TERM_PROGRAM is
+  ;; set, and the path should already be okay.
+  (if (not (getenv "TERM_PROGRAM"))
+      ;; Ask the shell to print out its value for $PATH.
+      (let ((path (shell-command-to-string "$SHELL -cl \"printf %s \\\"\\\$PATH\\\"\"")))
+        ;; Set the PATH variable to this to match everything up.
+        (setenv "PATH" path)
```
cg505, line 50: Why would asking the shell to print `$PATH` give you anything different then just getting `$PATH` itself?

.emacs.d/lisp/system/darwin.el
```diff
@@ -0,0 +1,44 @@
...
+only the global value for the variable 'PATH'.
+
+As such, access to things like Homebrew-installed packages is
+restricted.  This function calls the shell and asks it to print
+out the 'PATH' variable after it loads the `.profile'-related
+stuff.  This also respects the user's shell decision."
+
+  ;; Only run this once.  After you perform this step, TERM_PROGRAM is
+  ;; set, and the path should already be okay.
+  (if (not (getenv "TERM_PROGRAM"))
```
cg505: How does this get set? Are we certain this won't get set if this function isn't run?

cg505 (changes requested):
Adding a new hub has quite a bit of overhead: I'm not convinced it's the right choice here. I would think it could fit in `behavior`. What other sort of stuff would you expect to put in `system`?

rye:
This gets its own special hub because I'm sure there's compatibility stuff we can check. I didn't put this in behavior because I view these configurations as vital to proper operation rather than supplemental tweaks to editor behavior—I treat OS consistency and compatibility as it's own meta-level (read: system-level) form of behavior rather than a behavior of the editor itself. That said, I'm open to just tossing this in with behavior.

.emacs.d/lisp/system/darwin.el
rye: For "security reasons," PATH is not set when you start Emacs on MacOS. As such, asking the shell to run effectively forces Emacs to act like it was started with PATH set.

.emacs.d/init.el
rye: Fix incoming soon.

.emacs.d/lisp/system/darwin.el
rye: Yeah, I can see that being smart. I knew we'd want something like that.

.emacs.d/lisp/system/darwin.el
rye: Ouch.

.emacs.d/lisp/system/darwin.el
rye: This gets set when you run `shell-command-as-string`; I believe that when these compatibility fixes are run, Emacs sets its own environment properly and this gets set. This won't get set unless it already is. I can't be sure though.

rye:
See [this StackOverflow answer](http://emacs.stackexchange.com/a/553) for more details as to why the PATH-related stuff is necessary.

Specifically,
>The problem on [macOS] is that [macOS] does not set the environment the same when you call a program from the global UI or when you call it from a shell.  This means that running Emacs from a shell will result in different environment variables being set than when you run it from Finder.  This is especially annoying if you set environment variables in `.bashrc` or similar, [e.g. `.profile`, `.zshrc`] as that won't affect the "global" Emacs.

It is worth noting that there exists a package, [`exec-path-from-shell`](https://github.com/purcell/exec-path-from-shell) which could do the same thing for us.  @cg505, for this, would you rather we add that as a dependency and do that?  I can do that on a branch from this branch.

rye:
It's worth mentioning that the entire motivation for PATH fixing is that `dired` complains about `ls` not being sane, because the BSD `ls` is non-POSIX-compliant, and so I, personally, installed `coreutils` via Homebrew and linked it, but the PATH variable didn't contain it.

I can confirm that the `exec-path-from-shell` package works to perform the same operation, but, as I mention in #24, introduces a large amount of overhead because it performs too much extra stuff.

rye:
I am still down to move this into `behavior` if we deem that appropriate.

cg505:
Thank you for the explanation regarding PATH. I agree that the overhead of an entirely new packages is unreasonable. Your approach seems fine. I'll look at the new changes.

.emacs.d/lisp/system/darwin.el
```diff
@@ -38,7 +47,7 @@ stuff.  This also respects the user's shell decision."
 
 (if (eq system-type 'darwin)
     (progn
-      (kotct/darwin-swap-command-option)
+      (kotct/darwin-maybe-swap-command-option)
```
cg505: Unless I'm mistaken, this will test the value of `kotct/darwin-swap-command-option` when it's run, so it'll always be `t`. If a user sets it in their personal config, that won't have any effect. I propose running `kotct/darwin-maybe-swap-command-option` on `after-init-hook` to make sure any changes to the variable in personal configs are accounted for.

cg505 (changes requested):
Except for that one thing, looks good.

rye:
How's this now?

.emacs.d/lisp/system/darwin.el
```diff
@@ -47,7 +53,7 @@ stuff.  This also respects the user's shell decision."
 
 (if (eq system-type 'darwin)
     (progn
-      (kotct/darwin-maybe-swap-command-option)
-      (kotct/darwin-load-path-from-shell)))
+      (add-hook 'after-init-hook 'kotct/darwin-maybe-swap-command-option)
+      (add-hook 'after-init-hook 'kotct/darwin-maybe-load-path-from-shell)))
```
cg505: funquote (`#'`) plz

cg505 (changes requested):

rye:
Now?

cg505 (approved):
😍

rye:
@cg505 if you would like to see this pulled out into `behavior`, we can do that sometime.

cg505:
I feel like your reasoning for keeping it separate is fine. If down the road we don't see any other major additions to the hub than we can merge it back in with some of the others.

cg505:
This actually isn't doing the trick for me. I didn't realize this was in place and so I installed exec-path-from-shell. I'll take another look tomorrow and figure out why it's not working for me.

cooperc0:
It's working fine, idk what I was thinking.

cg505:
It's working fine, idk what I was thinking.

PR #24, opened 2017-02-05
darwin: Use `exec-path-from-shell`

rye:
This is an improvement on the `darwin` branch and will change the behavior to use `exec-path-from-shell` instead of our rolled version.

The chief problem I see with this is that it slightly adds even more delay on startup.  With this package, the increase in init time is `2.2` seconds for me due to a bulky shell configuration.  The&mdash;arguably much leaner&mdash;version on the `darwin` branch only increases my init time by `0.7` seconds.

It is worth noting that my actual shell's init time is, in these tests, nominally anywhere between `0.5` and `1.5` seconds.  So, the code on the `darwin` branch is much, ***much*** faster than the code on this branch.  This PR is only submitted for comparison, and I *strongly* recommend against including this with the changes on `darwin`.


rye:
In case you're wondering if just copying `PATH` is any faster, it is, but only marginally-so. (Changing results in a `0.4` second reduction off the `2.2` second increase, making for a `1.8` second increase.)

PR #25, opened 2017-02-10
Always delete trailing whitespace before saving

rye:
It's gross to have trailing whitespace.  In fact, most VC systems whine about it.  What's the point of having it?

We now just call `delete-trailing-whitespace` on `before-save-hook`.


cg505 (approved):
Looks good. Random other question: do we always ensure we have a final newline?

rye:
Nope.  I should do that, too, but in a separate pull request.

rye:
I'll go ahead and merge.

PR #27, opened 2017-02-28
`goto-line` Keybind

rye:
I've wanted this for some time now.  RFC?  Would this be better put at a different location?


rye:
Merging without review.

PR #31, opened 2017-03-02
Consolidate ace-jump-mode configurations

### LINKED ISSUE #29: ace-jump split between files?
There's ace-jump stuff in both `behavior/text.el` and `behavior/ace-jump.el`. We need to consolidate. Personally, I think we should go for `text.el`.

rye:
Yet another trivial change.

In general, I actually much favor having stuff split up into more files, especially for things that require their own configuration.  But not if it's only a few lines of configuration.

*Closes #29.*


rye:
Merging without review because this is not a big change.

PR #35, opened 2017-03-30
Don't require Git from user/user-config-system

### LINKED ISSUE #34: Emacs (25.1.1) Fails on Startup
In CI, Emacs is failing to start up in certain cases due to [line 3 in `user-config-system.el` as of](https://github.com/kotct/dot/blob/5ad51d14ee28a367d55db1e76681f0ce18d18b44/.emacs.d/lisp/user/user-config-system.el#L3) 5ad51d14ee28a367d55db1e76681f0ce18d18b44.

This issue existed prior to this commit as well, but I'm too lazy to bisect.

Since it looks like we don't even use it in that file, I think the fix is to remove that line and *not* require `git`.

rye:
Since it was unused, there's no reason to load it.  The only other solution to #34 that I can think of is to make this `require` optional.

*Closes #34*.

### DIFF
--- a/.emacs.d/lisp/user/user-config-system.el
+++ b/.emacs.d/lisp/user/user-config-system.el
@@ -1,7 +1,5 @@
 ;;; C-x C-z: switch personal configs
 
-(require 'git)
-
 (defvar kotct/user-current-username
   nil
   "The username associated with the currently loaded personal config.")


rye:
I'm going to merge this without review since #34 is a breaking issue for me.

PR #36, opened 2017-03-31
Custom Font Rescaling System

rye:
The default, built-in Emacs font re-scaling system [has](http://unix.stackexchange.com/q/7507/202251) [some](http://unix.stackexchange.com/q/29786/202251) [issues](https://github.com/syl20bnr/spacemacs/issues/1647).

This is a fix which is more of a re-implementation of the built-in Emacs font rescaling keybindings.  However, it adds functionality by allowing users to set a default font size (configured via `kotct/font-default-height`) and the factor by which they want to scale when they use the scaling system.

When scaling is applied via `C-x C-+`, `C-x C-=`, or `C-x C--`, the call to `kotct/font-rescale-height` is given either `+1` or `-1`. `kotct/font-rescale-factor` is then multiplied by this number to produce the change in height.

This works quite well for me.

[`rye-hub` currently contains](https://github.com/rye/.emacs/blob/67935cf788cf7938ba1e370c797df90d53685dec/rye-hub.el#L6-L7) an additional key-binding which makes this quite useful for jumping to a specific font size.

Yes, [a fix][linum-fix-using-builtin] does exist which uses the built-in rescaling system, but since it's much more complicated and this option provides more features and configuration for users to use, I think this is better.

[linum-fix-using-builtin]: http://unix.stackexchange.com/a/30087/202251

*This PR is held until review by @cg505.*

### DIFF
--- a/.emacs.d/lisp/visual/text-visuals.el
+++ b/.emacs.d/lisp/visual/text-visuals.el
@@ -10,14 +10,14 @@
 ;; Minor mode which highlights TODO, FIXME, DONE, FAIL, etc.
 (add-hook 'prog-mode-hook 'hl-todo-mode)
 
+(defvar kotct/font-default-height
+  12.0
+  "The default font height.")
+
 ;; Sets the font name to NAME.
 (defun kotct/font-set-name (name)
   "Sets the `:font' face attribute on the `default' face to match NAME."
   (set-face-attribute 'default nil :font name))
-
-(defvar kotct/font-default-height
-  12.0
-  "The default font height.")
 
 ;; Sets the font height to what a user would expect their value to map to.
 ;;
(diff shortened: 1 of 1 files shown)


cg505:
One thing to be aware of is that this increases the size of the entire frame, not just the buffer. I'd like to have the original system still available because of this reason. I'm not too worried about the patch @rye mentioned.

Could we provide the original functions in keybindings prefixed by `C-c`?

rye:
> [...] the patch @rye mentioned.

@cg505, do you mean [this StackOverflow answer](https://unix.stackexchange.com/questions/29786/font-size-issues-with-emacs-in-linum-mode/30087#30087)?  I'm all for adding more code if it makes for more robust behavior.  Or maybe did you mean [the added keybind I have in `rye-hub`](https://github.com/rye/.emacs/blob/646a6d6/rye-hub.el#L8-L9)?

> Could we provide the original functions in keybindings prefixed by `C-c`?

Probably. These are bound to functions in `face-remap.el`. See [Text Scale](https://www.gnu.org/software/emacs/manual/html_node/emacs/Text-Scale.html) in the Emacs Manual for details.

Note that all three of the keys `C-x C-{+,=}`, `C-x C--`, and `C-x C-0` are bound to `text-scale-adjust`, which, according to the manual,

> looks at the last key typed to determine which action to take.

So it should be sufficient to unbind and then re-bind each of the four key-bindings with a `C-c` prefix.  I think this should handle this behavior.

I will open an issue for this refinement.

cg505:
yeah, I meant the stackexchange answer

.emacs.d/lisp/visual/text-visuals.el
```diff
@@ -29,10 +34,35 @@
...
+    (set-face-attribute 'default nil :height (floor (* 10.0 new-size)))))
+
+(defun kotct/font-upscale-height () (interactive "*") (kotct/font-rescale-height +1.0))
+(defun kotct/font-downscale-height () (interactive "*") (kotct/font-rescale-height -1.0))
+(defun kotct/font-reset-height () (interactive "*") (kotct/font-set-height kotct/font-default-height))
+
+(global-unset-key (kbd "C-x C-+"))
+(global-unset-key (kbd "C-x C-="))
+(global-unset-key (kbd "C-x C--"))
+(global-unset-key (kbd "C-x C-0"))
```
cg505, line 55: are you sure these lines are needed?

rye: They probably aren't, to be honest. I'll rework this later this evening after I'm done working through some of my schoolwork.

PR #39, opened 2017-05-07
fixes #38; error in emacs-nox

### LINKED ISSUE #38: undefined function causes emacs config to fail in emacs-nox
`scroll-bar-mode` is undefined in emacs-nox. this causes the emacs config to fail.

samontea:
@rye check this. i doubt there will be any problems

### DIFF
--- a/.emacs.d/lisp/visual/frame-c.el
+++ b/.emacs.d/lisp/visual/frame-c.el
@@ -1,6 +1,7 @@
 ;; disable tool bar and menu bar
 (tool-bar-mode -1)
 (menu-bar-mode -1)
-(scroll-bar-mode -1)
+(if (boundp 'scroll-bar-mode)
+    (scroll-bar-mode -1))
 
 (provide 'frame-c)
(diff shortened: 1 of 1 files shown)


.emacs.d/lisp/visual/frame-c.el
```diff
@@ -1,6 +1,7 @@
 ;; disable tool bar and menu bar
 (tool-bar-mode -1)
 (menu-bar-mode -1)
-(scroll-bar-mode -1)
+(if (boundp 'scroll-bar-mode)
```
rye: Should this be `fboundp`?

samontea: yes according to our style guides, but it looks like there's not a difference. http://www.gnu.org/software/emacs/manual/html_node/elisp/Function-Cells.html

rye: Would you mind adding the f just so that we're maintaining the stuff?

samontea: yeah. googling `fboundp` was all I could do at the gym. 😘

rye (approved):
Excellent. LGTM.

cg505:
babe funquote that shit

rye:
Oh shoot, I missed that.

cg505:
nbd

PR #40, opened 2017-05-08
frame-c: Use a function quote for fboundp conditional

rye:
This is a quick improvement on #39.


cg505:
if you've tested it, looks good

PR #41, opened 2017-05-08
Don't fail horribly when we are missing packages and starting headless

### LINKED ISSUE #21: Startup Prompts Don't Work When Started Headless
We do several prompts on startup (especially the first time) such as asking if packages want to be installed, asking about a default config, etc. These currently don't work when emacs starts headless (e.g. `$ emacsclient -c -a ""`).

This is not a simple issue to solve, and thus is pretty low-priority. We would have to do something like stop startup, and put all remaining starup on `after-make-frame-functions`, and remove the hook it after it runs.

cg505:
My approach is what I initially suggested on #21: when we find out we're headless (i.e. `emacs --daemon`) and we don't have everything we need, halt startup and wait until we have an actual frame to continue. This is accomplished by implementing the additional function `kotct/load-hubs` and using throw-catch from `kotct/check-dependency-list` to `init.el`.

This fix works both for global dependencies and for user-specific dependencies.

I realized as I was typing this PR that I need to do a little bit of clean-up: one more commit is incoming.

*Closes #21*.

### DIFF
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -57,7 +57,11 @@ If AUTOLOADS is non-nil, update the autoloads for that directory."
   (normal-top-level-add-to-load-path kotct/hub-list))
 
 (defun kotct/load-hubs (&optional frame)
-  ;; (if ignore (error "%s" (frame-terminal ignore)))
+  "Does the majority of initialization for dot/.emacs,
+by loading the hubs defined in `kotct/hub-list'.
+
+Pass FRAME if the function is being called on `after-make-frame-functions'
+and we need to remove the hook and specifically use the frame."
   ;; If byte-compiled files are older, load newer version.
   (let ((load-prefer-newer t))
     (when frame
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -56,15 +56,28 @@ If AUTOLOADS is non-nil, update the autoloads for that directory."
   (add-to-list 'load-path default-directory)
   (normal-top-level-add-to-load-path kotct/hub-list))
 
-;; If byte-compiled files are older, load newer version.
-(let ((load-prefer-newer t))
-  ;; require all hubs
-  (mapc (lambda (hub)
-          (require (intern (concat hub "-hub"))))
-        kotct/hub-list))
+(defun kotct/load-hubs (&optional frame)
+  ;; (if ignore (error "%s" (frame-terminal ignore)))
+  ;; If byte-compiled files are older, load newer version.
+  (let ((load-prefer-newer t))
+    (when frame
+      (select-frame frame)
+      ;; second run, we already have the resources to verify packages
+      (kotct/check-dependency-list frame)
+      (remove-hook 'after-make-frame-functions #'kotct/load-hubs))
+    ;; require all hubs
+    (mapc (lambda (hub)
+            (require (intern (concat hub "-hub"))))
+          kotct/hub-list)
+    ;; Load the loaddefs.
+    (require 'kotct-loaddefs)))
 
-;; Load the loaddefs.
-(require 'kotct-loaddefs)
+;; if we are running in daemon-mode and some packages aren't installed,
+;; wait until a frame is created to finish loading
+(if (eq 'daemon-mode
+        (catch 'daemon-mode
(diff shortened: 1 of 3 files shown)


.emacs.d/lisp/package/verification.el
```diff
@@ -1,15 +1,25 @@
...
-          (require 'packup)
-          (kotct/packup-install-dependencies nil))))
+(defun kotct/check-dependency-list (&optional frame)
+  "Check to make sure all dependencies in `kotct/dependency-list' are installed.
+If not, interactively prompt the user to install them via `kotct/packup-install-dependencies'.
 
-(kotct/check-dependency-list)
+FRAME is used to determine whether or not we are running headless
+(e.g., as a result of `emacs --daemon`). If we are, throw 'daemon-mode.
+It is the caller's responsibility to catch this and handle it properly.
```
rye, line 10: How best to handle this?

.emacs.d/lisp/package/verification.el
```diff
@@ -1,15 +1,25 @@
...
+Typically this happens in init.el."
+  (if (not (every #'package-installed-p kotct/dependency-list))
+      (if (string= (terminal-name (frame-terminal frame)) "initial_terminal")
+          (throw 'daemon-mode 'daemon-mode)
+        (y-or-n-p "You don't seem to have all necessary packages installed. Install them?\n(Your emacs will probably not work if you don't.)")
```
rye: `s/Your emacs/This configuration/` or something of the like? Like technically you can "own" Emacs but meh.

.emacs.d/init.el
```diff
@@ -56,15 +56,32 @@ If AUTOLOADS is non-nil, update the autoloads for that directory."
...
-  ;; require all hubs
-  (mapc (lambda (hub)
-          (require (intern (concat hub "-hub"))))
-        kotct/hub-list))
-
-;; Load the loaddefs.
-(require 'kotct-loaddefs)
+(defun kotct/load-hubs (&optional frame)
+  "Does the majority of initialization for dot/.emacs,
+by loading the hubs defined in `kotct/hub-list'.
```
rye, line 61: Is the text justification correct here?

rye (commented):
Looks good but just a couple of questions/text tweaks.

rye:
I will do a quick functional check here as well. Switching to the branch, will report back.

.emacs.d/init.el
cg505: What qualifies as "correct"?

.emacs.d/lisp/package/verification.el
cg505: Yeah idc that sounds fine. That language is unchanged by the PR but I can update it.

.emacs.d/init.el
rye: Is this not supposed to be 80 columns?

rye:
FWIW, in the stead of a 2.3 second startup time, a headless daemon start on this branch takes 0.5 on my system, meaning a significant amount of speed improvement is taking place here.

Very nice work, @cg505. If you want me to, I can just approve this PR and we can ship this.

.emacs.d/lisp/package/verification.el
cg505: like done in init.el???

rye: Gotcha.

cg505:
That 0.5 startup time is a false positive since the startup is aborted, and the actual startup doesn't happen during that time. Still, fast. Thanks. I'm checking a few last things.

.emacs.d/init.el
cg505: lol idk

rye:
Right. But for all intents and purposes my reaction time after a frame is created is too slow to notice the second part of startup. I'm just saying that it feels much faster now.

.emacs.d/init.el
cg505: i just kind of do line breaks wherever i feel like

rye: ok. doesn't really matter that much. We can always go back and work on documentation and stuff later.

rye (approved):
This was really my only gripe so this PR LGTM now.

PR #43, opened 2017-05-11
Add machine-local configuration option

### LINKED ISSUE #30: Machine-Local Configuration
Provide some way to have a machine-specific config. Could be used for things like setting executable locations.

cg505:
Allows the user to add a machine-local config at `$XDG_CONFIG_HOME/dot/init.el`. If `$XDG_CONFIG_HOME` is not defined, it defaults to `~/.config/`.

*Closes #30*.

### DIFF
--- a/.emacs.d/lisp/user/machine-local.el
+++ b/.emacs.d/lisp/user/machine-local.el
@@ -1,6 +1,6 @@
 (defun kotct/dot-config-directory ()
   "Returns the directory name for where our machine-local Dot configuration directory should be."
-  (concat (file-name-directory (or (getenv "XDG_CONFIG_HOME") "~/.config")) "dot/"))
+  (concat (or (getenv "XDG_CONFIG_HOME") "~/.config/") "dot/"))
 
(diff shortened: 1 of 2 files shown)


.emacs.d/lisp/user/machine-local.el
```diff
@@ -0,0 +1,5 @@
+(let ((dot-config (concat (file-name-directory (or (getenv "XDG_CONFIG_HOME") "~/.config/")) "dot/"))
```
rye: Perhaps put this `let` in a `defun` or something like that?

rye: Just so that we are able to perhaps extend this functionality in the future, that is.

rye (commented):
LGTM, and is pretty light-weight, but maybe make this a defun?

cg505:
ok. can't get to it rn, so if you want to do it, go ahead

rye (approved):
This now looks good to me.  We might consider using `file-name-directory` a bit more semantically, but for now this looks great.

PR #46, opened 2017-05-29
recover from failed user configs

### LINKED ISSUE #7: kotct/user-switch-username Fails If Previous User Switch Failed
If I switch to a user configuration hub and then back to my own with a syntax error, things break and Emacs complains that, for example if I switch from my own broken hub to `cg505`'s, then try to switch to my own broken hub,

```
cg505-hub is not a currently loaded feature
```

samontea:
Without this change a failed user config causes `kotct/user-current-username` not to stay in sync with the current username which means all following calls to `kotct/user-switch-username` fail because of a failure in `kotct/user-unload-username`.
I believe this also resolves #7.

### DIFF
--- a/.emacs.d/lisp/user/user-config-system.el
+++ b/.emacs.d/lisp/user/user-config-system.el
@@ -124,7 +124,7 @@ If USERNAME is nil, prompt for a username."
   (message "Loading for username: %s" username)
   (condition-case err
       (kotct/user-load-username username)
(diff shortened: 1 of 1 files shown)


rye (approved):
LGTM.

cg505:
lmao i'm pretty sure i just fucked up when writing that code

samontea:
me2

PR #52, opened 2017-06-06
EditorConfig: Actually print warning if C Core not found

### LINKED ISSUE #51: Actually Print EditorConfig Warning Messages
[Editorconfig](http://editorconfig.org) overrides existing language code style settings in favor of project defined style guides in the form of `.editorconfig` file. This package does [require an external dependency for the emacs package](https://github.com/editorconfig/editorconfig-emacs#installation). We need to indicate to the user that they need to install this dep.

rye:
This is just a simple fix for #51.  I'm not sure if writing a message is the right way to do this, and if we wanted to use `lwarn`, we could, just changing `:debug` to `:warning`.  I think this follows our pattern for previous things.

*Closes #51.*

### DIFF
--- a/.emacs.d/lisp/code/editorconfig-c.el
+++ b/.emacs.d/lisp/code/editorconfig-c.el
@@ -23,7 +23,7 @@ Debug warning is suppressed if SUPPRESS is non-nil."
 
   (if (not suppress)
       (if (not (executable-find editorconfig-exec-path))
-          (lwarn '(editorconfig core) :debug "EditorConfig executable not found."))))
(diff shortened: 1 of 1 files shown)


samontea:
I don't like this solution, but don't know a better one. @cg505

rye:
Why don't you like it, then?

cg505:
I think the change in this PR is fine but we should revamp this whole file so there isn't much point to the PR. I'll have more detail when I get home.

rye:
Yes, I agree with this. To fix the issue with a message not getting displayed, I'm going to merge this. We need to come up with a better solution in general though.

Issue #57, opened 2017-06-07
Versioning


rye:
We should start thinking about versioning this using SemVer.  The two questions I can think of are:
* **Do we want to version this project?  Or should this be rolling-release?**
  * If so, **what will be our indication that we are ready to release version 1?**

This might affect our update-checking strategy for #56; users could also use a variable to decide whether they only want to use the latest release or the Git HEAD. (`master` in our case)

If we decide that we already passed a point where the project could be considered at v0.x.y, we can go back and make releases now for those points in history, and I can do that.

rye:
This will also allow us to use milestones going forward, which is nice.

cg505:
rolling-release bruh

rye:
>rolling-release bruh

While I like the idea of rolling-release, having some idea of stable vs development would be smart.  SemVer doesn't impose much overhead, and it's fun to release new versions.

Versioning is important especially to users who don't want to update every time their editor starts, and who are less likely to report and tolerate issues as they happen.  Milestones are a better way of structuring work, as well.

It's important to note that I have some people who are using this configuration here at St. Olaf, so it would make sense to keep `master` as a stable/development branch (tested on CI via #53) and create versions every so often.

There's certainly nothing stopping us from tagging previous commits and establishing an existing version structure, either.

cg505:
Okay, we can semver but also have a rolling-release system that works well?

rye:
What exactly do you mean by a rolling-release system?  I think it's totally fine to release new versions of dot as new stuff gets proven stable in master, so in that way we'd be rolling-release.  But we'd just do semver rather than having an incrementing number.  And `master` would always be the latest development version.

cg505:
So... here's my idea.

As far as semver: a semver breaking change in our context is something that might cause you to have to change something in your personal config. (This is not completely without exception, but should generally be followed.)

- `master` branch: Has all the latest changes, and should always be in a stable state. Will eventually become the next `minor-x.x` branch.
- `stable` branch: Has the most recent semver release. Should always mirror the latest `minor-x.x` branch.
- `minor-x.x` branches for each minor version that allow us to release hotfixes to old minor versions that people might still be using and don't want to have to update their config yet. (Branch off the oldest `minor-x.x` branch that we want to give the hotfix for, make the hotfix, merge the hotfix branch back into all `minor-x.x` branches that need the hotfix.)
[...]

PR #60, opened 2017-07-14
Add Versioning Guide to CONTRIBUTING file

rye:
This is a fairly straightforward implementation of the early spec reached in #57, which is still awaiting feedback from @samontea.

Please do provide feedback as you see fit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,4 +59,4 @@ In addition, we use the following scheme for additional branches:
 - `feature/` can be used as a prefix for feature branches, but this is not recommended.
   (Just use `kebab-case` for your branch names, we assume that anything without a special prefix is a normal feature PR.)
 - `hotfix/` should be used as a prefix for hotfix branches.
-  If a hotfix targets a release, (typical) a `release-vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release-vX.Y.Z` and the `hotfix/` branch.
+  If a hotfix targets a release, (which is the typical case) a `release-vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release-vX.Y.Z` and the `hotfix/` branch.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,21 +42,21 @@ A "breaking change" in our context is something that might cause a user to have 
 In general, our branching workflow consists of two persistent branches:
 
 - `master`, which contains all of the latest changes and should be in a stable state.
-  This is used as the base for all of the `release-` branches (see below.)
+  This is used as the base for all of the `release/*` branches (see below.)
 - `stable`, which is a tracking branch that gets releases merged into it.
   This is the branch to be used by users who only want stable features to be added to their configuration.
 
 In addition, we use the following scheme for additional branches:
 
-- `release-vX.Y` and `release-vX.Y.Z` branches are branches used to control versioning.
+- `release/vX.Y` and `release/vX.Y.Z` branches are branches used to control versioning.
   - To release a new version from `master`,
     1. after merging all PR's into `master`,
-    2. Create your `release-v` branch.
-    3. Bump the version on the `release-v` branch, either through a PR on a separate branch or something like that.
+    2. Create your `release/vX.Y(.Z)` branch.
+    3. Bump the version on the `release/vX.Y(.Z)` branch, either through a PR on a separate branch or something like that.
        If you have any release notes to add, do so here.
     4. Merge the branch back into `master` via a PR, including writing release notes and such.
     5. Create a release targeting the PR merge commit.
 - `feature/` can be used as a prefix for feature branches, but this is not recommended.
   (Just use `kebab-case` for your branch names, we assume that anything without a special prefix is a normal feature PR.)
 - `hotfix/` should be used as a prefix for hotfix branches.
-  If a hotfix targets a release, (which is the typical case) a `release-vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release-vX.Y.Z` and the `hotfix/` branch.
+  If a hotfix targets a release, (which is the typical case) a `release/vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release/vX.Y.Z` and the `hotfix/` branch.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,4 +60,5 @@ In addition, we use the following scheme for additional branches:
 - `feature/` can be used as a prefix for feature branches, but this is not recommended.
   (Just use `kebab-case` for your branch names, we assume that anything without a special prefix is a normal feature PR.)
 - `hotfix/` should be used as a prefix for hotfix branches.
-  If a hotfix targets a release, (which is the typical case) a `release/vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release/vX.Y.Z` and the `hotfix/` branch.
+  In the discussion of a problem, the maintainers will decide what version to release the hotfixes as and will create a `release/vX.Y.Z` branch for you to use as your base.
+  If a hotfix targets a release, (which is the typical case) the nascent `release/vX.Y.Z` branch should thus be used; submit a PR for the `hotfix/` branch into the `release/vX.Y.Z` branch.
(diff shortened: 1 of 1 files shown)


cg505:
I thought we decided to ditch the `release/` branches and just tag instead?

rye:
From [this comment](https://github.com/kotct/dot/issues/57#issuecomment-315226276) it sounded like 

>the `release-` branch can get discarded after the release is released and the branch is in `master`

Basically we'd keep them around for just a short time while bumping versions and potentially adding to changelogs

rye:
But we would still use tags, and the `release-` branches would get deleted after the release took place.

CONTRIBUTING.md
```diff
@@ -31,3 +31,32 @@ In general, here's a suggested workflow:
...
+    1. after merging all PR's into `master`,
+    2. Create your `release-v` branch.
+    3. Bump the version on the `release-v` branch, either through a PR on a separate branch or something like that.
+       If you have any release notes to add, do so here.
+    4. Merge the branch back into `master` via a PR, including writing release notes and such.
+    5. Create a release targeting the PR merge commit.
+- `feature/` can be used as a prefix for feature branches, but this is not recommended.
+  (Just use `kebab-case` for your branch names, we assume that anything without a special prefix is a normal feature PR.)
+- `hotfix/` should be used as a prefix for hotfix branches.
+  If a hotfix targets a release, (typical) a `release-vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release-vX.Y.Z` and the `hotfix/` branch.
```
cg505: typo: "typically"

CONTRIBUTING.md
```diff
@@ -31,3 +31,32 @@ In general, here's a suggested workflow:
...
+In general, our branching workflow consists of two persistent branches:
+
+- `master`, which contains all of the latest changes and should be in a stable state.
+  This is used as the base for all of the `release-` branches (see below.)
+- `stable`, which is a tracking branch that gets releases merged into it.
+  This is the branch to be used by users who only want stable features to be added to their configuration.
+
+In addition, we use the following scheme for additional branches:
+
+- `release-vX.Y` and `release-vX.Y.Z` branches are branches used to control versioning.
```
cg505: What is the point of the `release-vX.Y` branch? Seems unnecessary to me. The one possible use is leaving one around that always points to the latest release in that minor version, as a convenience. This mostly seems unnecessary though.

CONTRIBUTING.md
```diff
@@ -31,3 +31,32 @@ In general, here's a suggested workflow:
...
+  - To release a new version from `master`,
+    1. after merging all PR's into `master`,
+    2. Create your `release-v` branch.
+    3. Bump the version on the `release-v` branch, either through a PR on a separate branch or something like that.
+       If you have any release notes to add, do so here.
+    4. Merge the branch back into `master` via a PR, including writing release notes and such.
```
cg505: Imo the PR should target `stable`, and the tag is on that merge commit into `stable`. `stable` then can be merged back into master. It doesn't make sense to have active devel going on in master, make a release branch, and then merge that right back into `master`. The diff should be 0.

cg505 (changes requested):
I disagree with how the release process is put forward. Discuss as needed, but if substantial discussion must be had, move it to the issue instead of having it in the PR.

CONTRIBUTING.md
rye: Bumping a version is something that should probably not really be done by people submitting changes.  That's project-level work that should really just be done by the maintainers when they know what they want to merge.  We'd delete the branch, but people would be able to create the branch again if we needed to add something to a previous version in a patch release.

cg505: So I understand how you would use a `release-v2.1.0` branch, for instance, but what's the point of the `release-v2.1` branch?

Also, can we consider doing `release/` so that we don't have two different types of prefixes?

CONTRIBUTING.md
rye: I would agree if you're saying that any development work off of `stable` should take place and be directed back to the `stable` branch, but some people like @cg505 and myself would probably be working off of `master`.  Again, making a `release-vX.Y` branch is intended for work _related to the release process_, (incl. merging PR's) and nothing else.  The point of branching off of one of these branches is to have a nice place to merge back into, but I don't expect 3rd parties (or even contributors who are submitting their own changes) to properly take care of documenting their changes.

I don't understand what you mean by
>The diff should be 0.

cg505: Popping this back into the issue (#57) for more discussion.

CONTRIBUTING.md
rye: Further discussion should probably take place on #57.

CONTRIBUTING.md
```diff
@@ -31,3 +31,32 @@ In general, here's a suggested workflow:
...
+
+In addition, we use the following scheme for additional branches:
+
+- `release/vX.Y` and `release/vX.Y.Z` branches are branches used to control versioning.
+  - To release a new version from `master`,
+    1. after merging all PR's into `master`,
+    2. Create your `release/vX.Y(.Z)` branch.
+    3. Bump the version on the `release/vX.Y(.Z)` branch, either through a PR on a separate branch or something like that.
+       If you have any release notes to add, do so here.
+    4. Merge the branch back into `master` via a PR, including writing release notes and such.
```
cg505: We decided to merge into `stable` via a PR?

CONTRIBUTING.md
```diff
@@ -31,3 +31,32 @@ In general, here's a suggested workflow:
...
+    5. Create a release targeting the PR merge commit.
```
cg505: Add a step to merge the release back into `master`. No PR needed for this one, just do it.

CONTRIBUTING.md
cg505: Also we should probably get at least one approval (on the release notes and on the release) in order to do this merge. Should already be enforced with our current PR rules. (Don't worry about changing the actual text of the file here, this was just meant to be a note.)

CONTRIBUTING.md
```diff
@@ -31,3 +31,32 @@ In general, here's a suggested workflow:
...
+- `feature/` can be used as a prefix for feature branches, but this is not recommended.
+  (Just use `kebab-case` for your branch names, we assume that anything without a special prefix is a normal feature PR.)
+- `hotfix/` should be used as a prefix for hotfix branches.
+  If a hotfix targets a release, (which is the typical case) a `release/vX.Y.Z` branch should be used as the base of the PR instead; submit a PR for both the `release/vX.Y.Z` and the `hotfix/` branch.
```
cg505: >submit a PR for both the `release/vX.Y.Z` and the `hotfix/` branch.

What does it mean to submit a PR for the `hotfix/` branch? Aren't we working in the `hotfix/` branch? We PR `hotfix/*` into `release/vX.Y.Z` and then (if X.Y is the current minor version) PR `release/vX.Y.Z` into stable and tag. Right?

cg505 (changes requested):
Looking pretty good. A few small remaining issues.

CONTRIBUTING.md
rye: Note that I haven't updated this file at all since our discussion in #57, I was just looking to highlight the actual issues in this file.  Will fix shortly.

CONTRIBUTING.md
rye: Isn't `master` merge-locked?

CONTRIBUTING.md
rye: Yeah, that's right.  `hotfix` goes into `release`, `release` goes into `stable`, `stable` gets tagged, `stable` gets merged back to `master`.

CONTRIBUTING.md
rye: This means we have to have a reviewed PR before we can merge.  That's why I was thinking about doing it the other way; merge to `master` then merge to `stable`.

cg505: mmmm. Since an admin will be doing the release, we can just open a PR, then override checks and immediately merge I guess.

CONTRIBUTING.md
cg505: Okay. Maybe clarify this language to say what is being merged into what in those PRs.

CONTRIBUTING.md
rye: Yeah, it'd work that way.

CONTRIBUTING.md
rye: Yep.

cg505 (approved):
feel free to merge

rye:
Whee! Merging now.

Issue #53, opened 2017-06-06
Testing and CI


rye:
I think we should really look into improving some of the ways we interact with users.  We should try to deal with edge cases in a smarter way, and generally clean up our code.  How do we feel about using CI and tests?  Testing our existing code might evoke good change in our codebase.

#### Task List:

- [x] #61 Add testing framework and integration with Travis CI, using buttercup
- [x] #75 Add coverage support and reporting to Coveralls using undercover
- [ ] Write tests for existing code
- [ ] Add linting support, fix linting issues, and check linting in Travis CI

cg505:
I'm doing research rn on the best approach for testing. We also should get linting set up and put both of those into CI. Suggestions on CI are welcome. I'll update this at some point soon with a roadmap for testing/linting.

cg505:
I'm thinking we should use [buttercup](https://github.com/jorgenschaefer/emacs-buttercup) for tests.

cg505:
Linting will be a combo of [CEDET's checkdoc](http://cedet.sourceforge.net/checkdoc.shtml) and [elisp-lint](https://github.com/AdrieanKhisbe/elisp-lint).

cg505:
I'm gonna start work on this using these. lmk asap if you have an issue with that. planning on using travis

cg505:
I edited the initial comment to include a task list.

rye:
I'm not sure about using Coveralls out of necessity.  Do you know if [CodeClimate](https://codeclimate.com) supports EmacsLisp?

cg505:
I'm pretty sure it doesn't. I already have basic Coveralls support in the coverage-reporting branch. The overhead is pretty minimal. I'm not super satisfied with it at this point, which is why I haven't opened a PR, but at the same time, it's kinda a "why not" thing.

cg505:
Are we good with using undercover? Due to the way coverage libraries for emacs lisp are implemented, it will only show coverage for files that are at least a little tested... so the percentages may be off at first. Still should be a useful tool and I already pretty much have it implemented.

rye:
I'm open to whatever you have, so long as it works.

PR #61, opened 2017-07-16
Add Travis integration

cg505:
This PR adds basic integration for Travis CI. Merging it now will allow future PRs to write tests for their code.

### DIFF
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -3,6 +3,13 @@
 ;; Initialize the package repository.
 (package-initialize)
 
+;; Create variable to determine if we are in dev mode or not.
+;; (Set this in your user hub.)
+(defvar kotct/dev-mode
+  nil
+  "Whether or not kotct/dot emacs dev mode is enabled.
+Makes tests runnable and other dev-only goodies.")
+
 
 ;;; Autoload Configuration
 (setf generated-autoload-file "~/.emacs.d/lisp/kotct-loaddefs.el")
@@ -108,6 +115,10 @@ Despite this, your config appears to have loaded successfully.")
 (let ((loaddefs-buffer (get-buffer (file-name-nondirectory generated-autoload-file))))
   (if loaddefs-buffer
       (kill-buffer loaddefs-buffer)))
+
+;; If in dev mode, load the test runner.
+(when kotct/dev-mode
+  (require 'dot-tests "~/.emacs.d/test/lisp/dot-tests"))
 
 
 ;;; Customization File
--- a/.emacs.d/test/lisp/behavior/completion-c-test.el
+++ b/.emacs.d/test/lisp/behavior/completion-c-test.el
@@ -5,7 +5,7 @@
     :var (orig-mode)
 
     (before-all
-      ; record orignal setting for later
+      ; record original setting for later
       (setf orig-mode kotct/ido-current-menu-mode))
 
     (after-all
--- a/.emacs.d/test/lisp/behavior/completion-c-test.el
+++ b/.emacs.d/test/lisp/behavior/completion-c-test.el
@@ -1,0 +1,47 @@
+(kotct/load-corresponding)
+
+(describe "completion config"
+  (describe "ido menu mode"
+    :var (orig-mode)
+
+    (before-all
+      ; record orignal setting for later
+      (setf orig-mode kotct/ido-current-menu-mode))
+
+    (after-all
+      ; restore original setting
+      (setf kotct/ido-current-menu-mode orig-mode))
+
+    (before-each
+      (spy-on #'ido-grid-mode)
+      (spy-on #'ido-vertical-mode))
+
+    (describe "is unset correctly"
+      (it "when using grid mode"
+        (let ((kotct/ido-current-menu-mode 'grid))
+          (kotct/ido-unset-menu-mode))
+
+        (expect #'ido-grid-mode :to-have-been-called)
+        (expect (spy-calls-args-for #'ido-grid-mode 0) :to-equal '(-1)))
+
+      (it "when using vertical mode"
+        (let ((kotct/ido-current-menu-mode 'vertical))
+          (kotct/ido-unset-menu-mode))
+
+        (expect #'ido-vertical-mode :to-have-been-called)
+        (expect (spy-calls-args-for #'ido-vertical-mode 0) :to-equal '(-1))))
+
+    (describe "is set correctly"
+      (it "to grid mode"
+        (let ((kotct/ido-current-menu-mode 'normal))
+          (kotct/ido-set-menu-mode 'grid))
+
+        (expect #'ido-grid-mode :to-have-been-called)
+        (expect (spy-calls-args-for #'ido-grid-mode 0) :to-equal '(+1)))
+
+      (it "to vertical mode"
+        (let ((kotct/ido-current-menu-mode 'normal))
+          (kotct/ido-set-menu-mode 'vertical))
+
+        (expect #'ido-vertical-mode :to-have-been-called)
+        (expect (spy-calls-args-for #'ido-vertical-mode 0) :to-equal '(+1))))))
--- a/.emacs.d/test/lisp/dot-tests.el
+++ b/.emacs.d/test/lisp/dot-tests.el
@@ -1,9 +1,22 @@
 (require 'buttercup)
 
+(defun kotct/quietly-load-file (filename &optional throw-error)
+  "Load FILENAME, but do not send a message on success.
+If loading throws an error, send a message with FILENAME and the error.
+If THROW-ERROR is non-nil, re-throw the caught error."
+  (condition-case err
+        (load filename nil 'nomessage)
+      (error
+       (message (format "Error while loading %s:\n%s" file (error-message-string err)))
+       (if throw-error
+           (error (cdr err))))))
+
 (defun kotct/load-corresponding ()
+  "Load the file corresponding to the current test.
+Only works when called in script mode."
   (when noninteractive
     (assert (string-match-p "-test\\.el$" load-file-name))
-    (load-file
+    (kotct/quietly-load-file
      (replace-regexp-in-string
       "-test\\.el$" ".el"
       (replace-regexp-in-string
@@ -11,8 +24,13 @@
        load-file-name)))))
 
 (defun kotct/run-tests ()
+  "Run all emacs lisp tests for the kotct/dot emacs configuration."
+  (interactive)
   (setf buttercup-suites nil)
-  (load "~/.emacs.d/test/lisp/behavior/completion-c-test.el" nil t)
+  (dolist (file (directory-files-recursively
+                 (file-name-directory (symbol-file #'kotct/run-tests))
+                 "-test\\.el\\'"))
+    (kotct/quietly-load-file file))
   (let ((buttercup-reporter
          (lambda (event arg)
            (condition-case err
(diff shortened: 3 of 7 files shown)


cg505:
This is a part of #53. (Does not fix.)

cg505:
Gonna squash those last two commits.

.emacs.d/test/lisp/behavior/completion-c-test.el
```diff
@@ -0,0 +1,47 @@
+(kotct/load-corresponding)
+
+(describe "completion config"
+  (describe "ido menu mode"
+    :var (orig-mode)
+
+    (before-all
+      ; record orignal setting for later
```
rye: typo: `orignal`

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+before_install:
+  - export PATH="/home/travis/.evm/bin:$PATH"
+  - git clone https://github.com/rejeep/evm.git /home/travis/.evm
+  - evm config path /tmp
+  - evm install $EVM_EMACS --use --skip
+  - ln -s "$(pwd)/.emacs.d" ~/.emacs.d
+install:
+  - emacs --script .emacs.d/test/lisp/install.el
+script:
+  - emacs --version
```
rye: This shouldn't be included in the `script` block; it's not part of our tests.

.emacs.d/test/lisp/dot-tests.el
```diff
@@ -0,0 +1,27 @@
...
+    (load-file
+     (replace-regexp-in-string
+      "-test\\.el$" ".el"
+      (replace-regexp-in-string
+       (regexp-quote "/test/") "/"
+       load-file-name)))))
+
+(defun kotct/run-tests ()
+  (setf buttercup-suites nil)
+  (load "~/.emacs.d/test/lisp/behavior/completion-c-test.el" nil t)
```
rye: Is `nil t` necessary here?  Can you extract the test running behavior out into a different function and call it a bunch of times in a master function rather than add more and more of these lines to this one function?  Or maybe do globbing to find all the tests?

.emacs.d/test/lisp/dot-tests.el
```diff
@@ -0,0 +1,27 @@
...
+  (when noninteractive
+    (assert (string-match-p "-test\\.el$" load-file-name))
+    (load-file
+     (replace-regexp-in-string
+      "-test\\.el$" ".el"
+      (replace-regexp-in-string
+       (regexp-quote "/test/") "/"
+       load-file-name)))))
+
+(defun kotct/run-tests ()
```
rye: Missing docstring?

.emacs.d/test/lisp/dot-tests.el
```diff
@@ -0,0 +1,27 @@
+(require 'buttercup)
+
+(defun kotct/load-corresponding ()
```
rye: Missing docstring?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+sudo: false
+cache: apt
+env:
+  - EVM_EMACS=emacs-25.1-travis
+  - EVM_EMACS=emacs-25.2-travis
+before_install:
+  - export PATH="/home/travis/.evm/bin:$PATH"
+  - git clone https://github.com/rejeep/evm.git /home/travis/.evm
+  - evm config path /tmp
+  - evm install $EVM_EMACS --use --skip
```
rye: This should move to the `install` block.

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: emacs-lisp
+sudo: false
+cache: apt
+env:
+  - EVM_EMACS=emacs-25.1-travis
+  - EVM_EMACS=emacs-25.2-travis
+before_install:
+  - export PATH="/home/travis/.evm/bin:$PATH"
+  - git clone https://github.com/rejeep/evm.git /home/travis/.evm
```
rye: Maybe consider cloning before exporting `PATH`?  It's just a matter of semantics.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - evm config path /tmp
+  - evm install $EVM_EMACS --use --skip
+  - ln -s "$(pwd)/.emacs.d" ~/.emacs.d
```
rye: Perhaps do this in the `install` block as well?

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: emacs-lisp
+sudo: false
+cache: apt
+env:
+  - EVM_EMACS=emacs-25.1-travis
+  - EVM_EMACS=emacs-25.2-travis
+before_install:
+  - export PATH="/home/travis/.evm/bin:$PATH"
+  - git clone https://github.com/rejeep/evm.git /home/travis/.evm
+  - evm config path /tmp
```
rye: Doesn't matter but maybe use `mktemp -d` to make a temporary directory specifically for the install?

rye (dismissed):
This is a good first step.  I just that some cleanup specifically related to the Travis config and some of the runners is needed to make this very refined.

.travis.yml
cg505: Don't think this really matters. EVM says to use `/tmp`, and I don't want to run into issues with that, so I'm just going to leave it.

.travis.yml
cg505: Do you advocate removing completely or moving to `before_script`?

rye: That would be smart, I think.

.emacs.d/test/lisp/dot-tests.el
cg505: `nil t` is necessary, otherwise we get a bunch of messages saying it loaded all the test files. It's annoying. I'm making this line dynamic so we don't have to specify all the test files.

rye: Do you prefer `nil t` over something more like `nil 'nomessage`?

cg505: That is fine.

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+install:
+  - evm install $EVM_EMACS --use --skip
+  - ln -s "$(pwd)/.emacs.d" ~/.emacs.d
+  - emacs --script .emacs.d/test/lisp/install.el
+  - echo "base-config" > .emacs.d/lisp/user/default-username
+before_script:
+  - emacs --version
+script:
+  - emacs --script ~/.emacs.d/init.el
```
rye: Is this needed and will it stay persistent to the next call?

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - emacs --script .emacs.d/test/lisp/run-tests.el
```
rye: Which one of these is needed?  Also, do you need `~/` at the start of this?

.travis.yml
```diff
@@ -0,0 +1,20 @@
+language: emacs-lisp
+sudo: false
+cache: apt
+env:
+  - EVM_EMACS=emacs-25.1-travis
+  - EVM_EMACS=emacs-25.2-travis
+before_install:
+  - git clone https://github.com/rejeep/evm.git /home/travis/.evm
```
rye: Consider using `$HOME` instead; `HOME=/home/travis` [by default](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables).

rye: You could also just use `~`, which is equivalent to `$HOME`.

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - export PATH="/home/travis/.evm/bin:$PATH"
```
rye: Ditto here, use `$HOME` or `~`?

.travis.yml
```diff
@@ -0,0 +1,20 @@
+language: emacs-lisp
+sudo: false
+cache: apt
+env:
+  - EVM_EMACS=emacs-25.1-travis
+  - EVM_EMACS=emacs-25.2-travis
```
rye: Can we add support for latest `HEAD` Emacs, too?  Does EVM have development versions?

.emacs.d/init.el
```diff
@@ -109,6 +116,10 @@ Despite this, your config appears to have loaded successfully.")
   (if loaddefs-buffer
       (kill-buffer loaddefs-buffer)))
 
+;; If in dev mode, load the test runner.
+(when kotct/dev-mode
+  (require 'dot-tests "~/.emacs.d/test/lisp/dot-tests"))
```
rye, line 121: I'm hesitant to include this (both the var and the thing) in `init.el` because
- It adds calls to the `init` stream under normal non-dev circumstances, and
- Adding this to the source tree for the entire configuration means that we're tainting our actual configuration with testing stuff.

I think tests should be run in a separate Emacs instance from the existing one, if the developer is running one.  The ability to run `kotct/run-tests` within the normal running circumstances is somewhat sketchy to me.  Far better to run Emacs using a `--script` that also loads `init` than do this, I think.

rye (changes requested):
Progress, but I'm hesitant to keep `dev-mode` in `init.el`.

.travis.yml
cg505: EVM has an `emacs-git-snapshot-travis` build but it seems like maybe not the best plan to have our tests dependent on unstable behavior in the emacs git repo? It also seems unnecessary. On second thought I don't really care very much and this is unlikely to break anything. I'm happy to go with whatever you think is the best plan.

rye: I typically test my Ruby projects against `ruby`'s latest release(s) as well as the latest Git snapshot because I like to see if stuff works with the new API.  It's okay to omit it, though.

.emacs.d/init.el
cg505: >It adds calls to the `init` stream under normal non-dev circumstances

Hardly. I wouldn't consider a variable definition and an if-statement to be adding anything.

>Adding this to the source tree for the entire configuration means that we're tainting our actual configuration with testing stuff.

I mean... the tests do not run or actually do anything unless you explicitly run them, so the only "tainting" is the definition of a few functions. Also this only happens in dev mode.

>I think tests should be run in a separate Emacs instance from the existing one, if the developer is running one. The ability to run `kotct/run-tests` within the normal running circumstances is somewhat sketchy to me.

Officially, a separate Emacs instance is the correct way to run the tests. If `emacs --script ~/.emacs.d/test/lisp/run-tests.el` fails, the tests do not pass. That is how they're run in Travis. That said, having the ability to run the tests in your Emacs session is a convenience that makes it much easier to run/debug tests, and there's very little overhead.

.travis.yml
cg505: The first one is intended to make sure that Emacs can successfully start without error. The second runs the tests.

The `~/` is just meant to replicate Emacs' behavior by running init.el from the home directory instead of the clone directory.

cg505: Probably we can make the presence of `~/` consistent without impact, I was just explaining why it is what it is rn.

.travis.yml
cg505: This just tests startup. Nothing persists to the next call (except maybe the creation of some random things like the recentf file).

.emacs.d/init.el
rye: This will be a good enough first step, I guess.  We can go ahead and continue with the PR, and if it causes problems we can rethink.

rye (approved):
LGTM for now! 🎉  🥇

PR #62, opened 2017-07-20
bugfix: load custom file after setting it

cg505:
Currently we set the `custom-file` but never load it, kind of defeating the purpose.

Also, this is the first instance of our tests catching something! Yay!


rye (approved):
Woops, that's a good catch.

PR #63, opened 2017-07-20
RELEASE v0.0.0

cg505:
This is the first release of the current beta state of dot. This is essentially a test release for the newly agreed-upon versioning system discussed in #57.


VERSION
```diff
@@ -0,0 +1 @@
+0.0.0
```
rye: Newline at EOF?

rye (changes requested):

VERSION
cg505: fuck

rye (approved):
Good stuff.

PR #65, opened 2017-07-20
Add options to magit popups for GPG signing commits

cg505:
This adds the "=S" (--gpg-sign=) option to the magit-merge and magit-rebase popups. It also includes tests for the magit conf.


.emacs.d/lisp/code/magit-c.el
```diff
@@ -1,5 +1,11 @@
-(setf magit-push-always-verify nil)
-(setf magit-last-seen-setup-instructions "1.4.0")
+(with-eval-after-load 'magit
+  (let ((popups-to-add-sign '(magit-merge-popup
+                              magit-rebase-popup)))
+    (mapcar (lambda (popup)
+              (magit-define-popup-option
+                popup
+                ?S "Sign using gpg" "--gpg-sign=" #'magit-read-gpg-secret-key))
+            popups-to-add-sign)))
```
rye, line 8: Where did `magit-push-always-verify` and `magit-last-seen-setup-instructions` go?  If they are not still needed, let me know.

rye (approved):
Spare my minor concerns, I like this.  LGTM! 🎉   Looking at this locally, it seems to work well enough.

.emacs.d/lisp/code/magit-c.el
cg505: They are not still needed. I can try and dig through the magit repo to find where they confirm this but I'd rather not.

rye: From [the 2.10 changelog](https://github.com/magit/magit/blob/1cb2e127d592915bf2d2efb18bacc5c1940ade1b/Documentation/RelNotes/2.4.0.txt#L63-L66):

>* The option `magit-push-always-verify` was removed.  That was only a temporary kludge to keep users from shooting themselves in the foot.  This is no longer necessary because one now always sees where one is about to push to.

It also appears that `magit-last-seen-setup-instructions` suffered the same fate.

rye: Thus no change is necessary, and I'm glad they're gone now.

Issue #55, opened 2017-06-06
Support for Smartparens


cg505:
Eventual support for [smartparens](https://github.com/Fuco1/smartparens) is a must at some point. It should be on the radar.

cg505:
Anyone want to take this? I can take it once I finish the testing framework if no one else wants to.

rye:
Sure!

cg505:
assign urself!

rye:
Done.

rye:
(With assignment.)

rye:
I have picked this up again.  @kotct/emacs, should we turn smartparens on globally?

samontea:
lol probably off by default it could be confusing for new users.

rye:
I was actually more asking whether or not it should be turned on for all languages that it supports or if I should manually add a hook for each language.  I'd prefer the former if it's workable.

cg505:
smartparens-like behavior is standard in most IDEs. I don't think it would be confusing or problematic to have smartparens enabled globally by default.

rye:
I agree, that's why I committed it like that. We can always throw in a variable to turn that global setting off.

PR #70, opened 2017-07-31
Add Smartparens support

rye:
For now, `smartparens-global-mode` is enabled and we are using the vanilla configuration provided by `smartparens`.  I'm open to feedback to make this work better for people and also to add new features, but bear in mind that these things can be added later on as we see the need for them.

### DIFF
--- a/.emacs.d/lisp/package/dependencies.el
+++ b/.emacs.d/lisp/package/dependencies.el
@@ -20,6 +20,7 @@
     elixir-mode ;; for editing Elixir code
     rust-mode ;; for editing Rust code
     buttercup ;; for tests
+    smartparens ;; for dealing with paired control flow symbols
 
     ;; THEMES
     solarized-theme)
--- a/.emacs.d/lisp/code/code-hub.el
+++ b/.emacs.d/lisp/code/code-hub.el
@@ -1,6 +1,7 @@
 (kotct/hub "code"
            (editorconfig-c
             magit-c
+            smartparens-c
             indentation
             code-navigation))
 
--- a/.emacs.d/lisp/code/smartparens-c.el
+++ b/.emacs.d/lisp/code/smartparens-c.el
@@ -1,3 +1,5 @@
 (require 'smartparens-config)
 
+(smartparens-global-mode t)
+
 (provide 'smartparens-c)
--- a/.emacs.d/lisp/code/smartparens-c.el
+++ b/.emacs.d/lisp/code/smartparens-c.el
@@ -1,0 +1,3 @@
+(require 'smartparens-config)
+
+(provide 'smartparens-c)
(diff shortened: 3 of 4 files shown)


cg505 (dismissed):
Add tests please (just test that when you load the file smartparens-global-mode gets activated).

Also, we should go through any languages that dot has preliminary support for and make sure smartparens is working okay. Seem pretty good out of the box to me but we still ought to check. This might not be part of this PR or #55 but just wanted to throw it on your radar.

rye:
>Add tests please (just test that when you load the file `smartparens-global-mode` gets activated).

This test will be a good first exposure to tests—I'll take a stab at it and request review again.

>Also, we should go through any languages that dot has preliminary support for and make sure smartparens is working okay. Seem pretty good out of the box to me but we still ought to check. This might not be part of this PR or #55 but just wanted to throw it on your radar.

For what it's worth, this has been tested fairly extensively (~15 minutes of checkride editing) with Ruby, HTML, JS, and CSS within `web-mode`, and of course, EmacsLisp.  Feel free to checkout and verify that this is the case for you, and if there's any weird stuff we can toss that in this first-pass configuration.  This is just the first step for this feature but will accomplish >90% of what we want to do.

Smartparens is freakishly extensible so if we want to add a custom tweak to this later, I'm all for it.

rye:
The one caveat that I noticed was that if you delete a line containing a paren, the matching paren does not get deleted.  Smartparens, of course, has a solution for this:
>You may want to try `smartparens-strict-mode`. This enforces that pairs are always balanced, so commands like `kill-line` keep your code well-formed.

(From the [Smartparens README](https://github.com/Fuco1/smartparens/blob/f493fdba6af036497e0f0a204c7f7388888bf6b3/README.md#getting-started).)

cg505:
I enable strict mode personally but it is VERY confusing if you don't understand what is going on and should NOT be enabled by default

rye:
I agree—it's a bit overbearing and can cause problems if you are editing in a weird way.  It might be something that we can document or write an enabler for, but besides that it's probably best if people just turn it on themselves.

.emacs.d/lisp/package/dependencies.el
```diff
@@ -20,6 +20,7 @@
     elixir-mode ;; for editing Elixir code
     rust-mode ;; for editing Rust code
     buttercup ;; for tests
+    smartparens ;; for dealing with paired control flow symbols
```
cg505, line 23: "paired control flow symbols" ok mr scientist lmao

cg505 (approved):
not really much to complain about here, looks good

.emacs.d/lisp/package/dependencies.el
rye: lmao

PR #75, opened 2017-08-03
Add Coverage reporting to coveralls.io

cg505:
https://coveralls.io/github/kotct/dot

Also add badges for travis and coveralls to the readme.

This is a part of #53.


cg505:
what

cg505:
@rye @samontea actually ready for review, sorry for the confusion

README.md
```diff
@@ -1,5 +1,7 @@
 # kotct/dot
 
+[![Build Status](https://travis-ci.org/kotct/dot.svg?branch=master)](https://travis-ci.org/kotct/dot) [![Coverage Status](https://coveralls.io/repos/github/kotct/dot/badge.svg)](https://coveralls.io/github/kotct/dot)
```
rye: Can these go on the same line as the title?

.travis.yml
```diff
@@ -17,8 +17,8 @@ install:
 before_script:
 - emacs --version
 script:
-- emacs --script .emacs.d/init.el
 - emacs --script .emacs.d/test/lisp/run-tests.el
+- emacs --script ~/.emacs.d/init.el
```
rye: Suggest changing both to `~/`.

.travis.yml
```diff
@@ -17,8 +17,8 @@ install:
 before_script:
 - emacs --version
 script:
-- emacs --script .emacs.d/init.el
 - emacs --script .emacs.d/test/lisp/run-tests.el
+- emacs --script .emacs.d/init.el
```
rye, line 21: Use `~/` for both?

.emacs.d/test/lisp/run-tests.el
```diff
@@ -4,4 +4,17 @@
 
 (load-file (concat (file-name-directory load-file-name) "/dot-tests.el"))
 
+(when (require 'undercover nil 'noerror)
+  ;; undercover can't search for matching files recursively
+  ;; override a local function that fixes this
+  (advice-add
+   #'undercover--wildcards-to-files :override
+   (lambda (wildcards)
```
rye: Isn't `wildcards` unused?

.emacs.d/test/lisp/install.el
```diff
@@ -4,4 +4,7 @@
 (package-initialize)
 (package-refresh-contents)
 
+;; add undercover for coverage reporting
+(setf kotct/dependency-list (cons 'undercover kotct/dependency-list))
```
rye, line 8: Erm.  Any particular reason for this?

rye (changes requested):

.travis.yml
cg505: We had a discussion about this already, let me find it.

.emacs.d/test/lisp/run-tests.el
cg505: I'm overriding a function that takes an argument so I need to allow an argument to be passed.

.emacs.d/test/lisp/install.el
cg505: So undercover is installed, but only on Travis (via install script instead of on init).

.travis.yml
cg505: https://github.com/kotct/dot/pull/61#discussion_r127622762

.emacs.d/test/lisp/run-tests.el
rye: Gotcha, does the argument need to be the same though?  (Like, can you do `(lambda (_)` on this line instead?  Or `(_wildcards)`?)  I haven't tested this but we might get lexical binding errors, see [EL manual § 11.9.4 "Using Lexical Binding"](https://www.gnu.org/software/emacs/manual/html_node/elisp/Using-Lexical-Binding.html):
> (To silence byte-compiler warnings about unused variables, just use a variable name that start with an underscore. The byte-compiler interprets this as an indication that this is a variable known not to be used.)

If it's not used, changing it to underscore would be nice.

.emacs.d/test/lisp/install.el
rye: Oh, I see.

.travis.yml
rye: Oh, my apologies, I thought there was a `~/` at the start of this path and I was very confused.

.emacs.d/test/lisp/run-tests.el
cg505: We don't use lexical binding in our code, so there's no way for emacs to know if a variable is unused. That said, it probably would be good style to put the underscore anyway.

rye: Well, style was what I was more concerned about anyways.

rye (approved):

Issue #66, opened 2017-07-29
Packup UI update


samontea:
Making the `packup` buffer interactive to allow for the user to more easily select a subset of the packages that they wish to update.

It would also help us prepare for the work necessary to complete #56. This can be considered party of that issue.

#### Acceptance Criteria:
- Must allow for user to select packages to update w/i a read only buffer
- Must provide help instructing users how to interact w/ the read only buffer

cg505:
dis gon be sick

cg505:
@samontea Close this issue?

Leaving a couple notes for future reference: support `n` and `p` like magit/dired. recursive dependency resolution?

PR #84, opened 2017-08-24
kotct/packup ui 

samontea:
implemented `kotct/packup` from #66 
- performed manual user testing.
- does not include unit testing

### DIFF
--- a/.emacs.d/lisp/package/packup.el
+++ b/.emacs.d/lisp/package/packup.el
@@ -15,6 +15,37 @@ Does not automatically refresh package list."
   (every (lambda (x) (package-installed-p package (package-desc-version x)))
          (cdr (assq package package-archive-contents))))
 
+(defun kotct/insert-package-row (package-name package-update-date package-desc-version)
+  "Inserts a package row into current buffer."
+  (let ((inhibit-read-only t))
+    (insert (format "[x] %s %s %s\n" package-name package-update-date package-desc-version))))
+
+(defun kotct/packup-initialize-buffer ()
+  "Initializes the packup buffer."
+  (kill-all-local-variables)
+;;  (use-local-map packup-mode-map) ;; mapppppps
+  (setq major-mode 'packup-mode
+        mode-name "Packup"
+        buffer-read-only t ;; read only
+        mode-line-buffer-identification "packup")
+  (package-refresh-contents)
+  (let ((install-list nil))
+    (dolist (package kotct/dependency-list)
+      (let ()
+        (if (or (not (package-installed-p package))
+                (not (kotct/package-up-to-date-p package)))
+            (apply 'kotct/insert-package-row (cons package (package-desc-version (cadr (assq package package-alist))))))))))
+
+;;;###autoload
+(defun kotct/packup ()
+  "Creates an interactive buffer to install/update packages."
+  (interactive)
+  (let ((old-buf (current-buffer))
+        (buffer (create-file-buffer "*packup*"))) ;; this is wrong...
+    (set-buffer buffer)
+    (kotct/packup-initialize-buffer)
+    (pop-to-buffer-same-window buffer)))
+
 ;;;###autoload
 (defun kotct/packup-install-dependencies (no-refresh &optional update)
   "Installs the dependencies.
--- a/.emacs.d/lisp/package/packup.el
+++ b/.emacs.d/lisp/package/packup.el
@@ -11,8 +11,7 @@
 
 (defvar kotct/packup-marker-char ?x
   "In Packup, the current mark character.
-This is what the do-commands look for, and the flag the mark-commands store."
-  )
+This is what the do-commands look for, and the flag the mark-commands store.")
 
 (defun kotct/package-up-to-date-p (package)
   "Returns true if PACKAGE is up-to-date.
@@ -65,7 +64,6 @@ Does not automatically refresh package list."
              (while next-position
                (goto-char next-position)
                (setq results (cons ,body results))
-               (message "hi")
                ;; move after last match
                (goto-char next-position)
                (forward-line 1)
@@ -87,8 +85,7 @@ Returns \"\" if there is no package name on the line."
            (if (re-search-forward " " nil t)
                (prog2
                    (backward-char)
-                   (buffer-substring p (point-marker))
-                   )
+                   (buffer-substring p (point-marker)))
              ""))))
 
 (defun kotct/packup-do-update ()
--- a/.emacs.d/lisp/package/packup.el
+++ b/.emacs.d/lisp/package/packup.el
@@ -230,7 +230,7 @@ If AUTO-UPDATE is non-nil, out-of-date/uninstalled packages will be updated."
                    (princ (symbol-name (car package)))
                    (princ (if (cdr package) " (update)" " (install)"))))
                (if (or auto-update
-                       (y-or-n-p "Auto install/update these package?"))
+                       (y-or-n-p "Auto install/update these package(s)?"))
                    (progn (package-download-transaction (package-compute-transaction () install-list))
                           (kill-buffer "*packup: packages to upgrade*")
                           (message "Dependency installation completed."))
--- a/.emacs.d/lisp/package/packup.el
+++ b/.emacs.d/lisp/package/packup.el
@@ -25,8 +25,7 @@ Does not automatically refresh package list."
   (let ((inhibit-read-only t))
     (insert (format "[%c] %s -> %s\n" kotct/packup-marker-char package-name package-desc-version))))
 
-(defun kotct/packup-mark-files-in-region (start end)
-  "" ;; TODO
+(defun kotct/packup-mark-packages-in-region (start end)
   (let ((inhibit-read-only t))
     (if (> start end)
         (error "start > end"))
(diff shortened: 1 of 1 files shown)


rye:
This is a great start—thanks for your work so far on this, @samontea! 🎉 💯

Some off-hand notes from my initial evaluation, which aren't really criticisms of this work but more general ideas of where we can go from this:

- After checking out and restarting Emacs, `M-x kotct/packup` fails with

  ```
  command-execute: Autoloading file /Users/kristofer/Git/kotct/dot/.emacs.d/lisp/package/packup.elc failed to define function kotct/packup
  ```
  when I switch branches, but this seems unrelated as deleting the `.elc` file for `packup.el` resolves this issue.
- Can we add `U` and `M` to the help if they have been implemented?  (It seems that they have been.)
- When there are no updates to perform, can the `packup` buffer say something and not be blank?  This could be a static idle thing defined in a variable.
- I don't know how `package-list` from Emacs does column headers, but do we want to have column headers?  Those are pretty cool.  Also, can we separate the version with `.` rather than ` `?  (Also, colors?)
- After executing, a buffer pops up listing everything to be installed.  As the installation process proceeds, this list buffer stays open while Emacs blocks for this installation.  This is normal and probably part of the existing installation procedure.  I'm wondering if we can do the heavy work in a way that doesn't bring that up or at least lets us customize.  EDIT: I think this is part of the existing packup stuff, and has little to do with the UI.
- Using this UI, obsolete packages are kept around—do we want to handle cleaning up for our users?

I will do an actual code review shortly, but again, this looks pretty awesome.

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,197 @@
 
 (package-initialize)
 
+(defvar kotct/packup-marker-char ?x
+  "In Packup, the current mark character.
+This is what the do-commands look for, and the flag the mark-commands store."
+  )
```
rye: Move to previous line?  I don't think we do this anywhere else.

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,197 @@
...
+         (if t
+             results
+           nil))))
+
+(defun kotct/packup-get-package-name ()
+    "Gets package name on current line.
+Returns \"\" if there is no package name on the line."
+    (save-excursion
+      (beginning-of-line)
+      (let (p)
```
rye: What is `p`?  (Better name?)

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,197 @@
...
+             ;; can insert lines before the just found package,
+             ;; confusing us by finding the same marked package again
+             ;; and again and...
+             (setq next-position (and (re-search-forward regexp nil t)
+                                      (point-marker))
+                   found (not (null next-position)))
+             (while next-position
+               (goto-char next-position)
+               (setq results (cons ,body results))
+               (message "hi")
```
rye: Remove "hi"?

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,197 @@
...
+  "Refreshes packages in current buffer."
+  (interactive)
+  (if (eq major-mode 'packup-mode)
+    (let ((inhibit-read-only t))
+      (erase-buffer)
+      (kotct/packup-initialize-buffer-contents))))
+
+(defun kotct/packup-help ()
+  (interactive)
+  (message "g-refresh m-mark u-unmark x-execute ?-help"))
```
rye, line 172: [Add `U` and `M` here?](https://github.com/kotct/dot/pull/84#issuecomment-324787794)

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,197 @@
...
+    (define-key map "g" 'kotct/packup-refresh)
+    (define-key map "?" 'kotct/packup-help)
+    map))
+
+;;;###autoload
+(defun kotct/packup ()
+  "Creates an interactive buffer to install/update packages."
+  (interactive)
+  (let ((old-buf (current-buffer))
+        (buffer (create-file-buffer "*packup*"))) ;; this is wrong...
```
rye: Hmm, buffer has name `packup` for me with no cool asterisks.  We should figure out how.  Here is `list-packages` from `package.el` in Emacs, which seems to use `get-buffer-create` instead.

```el
;;;###autoload
(defun list-packages (&optional no-fetch)
  "Display a list of packages.
This first fetches the updated list of packages before
displaying, unless a prefix argument NO-FETCH is specified.
The list is displayed in a buffer named `*Packages*'."
  (interactive "P")
  (require 'finder-inf nil t)
  ;; Initialize the package system if necessary.
  (unless package--initialized
    (package-initialize t))
  ;; Integrate the package-menu with updating the archives.
  (add-hook 'package--post-download-archives-hook
            #'package-menu--post-refresh)
  (add-hook 'package--post-download-archives-hook
            #'package-menu--mark-or-notify-upgrades 'append)

  ;; Generate the Package Menu.
  (let ((buf (get-buffer-create "*Packages*")))
    (with-current-buffer buf
      (package-menu-mode)

      ;; Fetch the remote list of packages.
      (unless no-fetch (package-menu-refresh))

      ;; If we're not async, this would be redundant.
      (when package-menu-async
        (package-menu--generate nil t)))
    ;; The package menu buffer has keybindings.  If the user types
    ;; `M-x list-packages', that suggests it should become current.
    (switch-to-buffer buf)))
```

rye (changes requested):
Very good, just a couple of tweaks to be made I think.

.emacs.d/lisp/package/packup.el
samontea: we're going to end up with a lot of commands bound. if we document everything in the help it's going to be insane. i think we should just let people run `describe-bindings`. this is the way that `dired` does things.

.emacs.d/lisp/package/packup.el
samontea: uhm you're straight tripping if you run `M-x kotct/packup` it creates a buffer that is called `*packup*`. i'm like sure of that

i used `create-file-buffer` because it was what `dired` uses. i agree i should have used `get-buffer-create`. pushing that up now.

rye: On Emacs HEAD it doesn't have asterisks, then, or they aren't displayed properly.

.emacs.d/lisp/package/packup.el
rye: That works.  Then bind `?` to `describe-bindings`, maybe?

.emacs.d/lisp/package/packup.el
samontea: i don't feel like it's ambiguous, and i don't feel like it deserves a better name. it's just a variable that's like not used and assigned two lines below to be a point. ¯\_(ツ)_/¯

.emacs.d/lisp/package/packup.el
samontea: uhm no that would be confusing to a new user. i think that we should do what `dired` does. if a user wants the keybindings described they can just run `C-h b`. we might investigate creating a control panel style help/command dialogue like `magit` has, but that's future work.

.emacs.d/lisp/package/packup.el
rye: Okay so like would `point` be better?

rye: Okay so like would `point` be better?

.emacs.d/lisp/package/packup.el
samontea: ¯\_(ツ)_/¯

samontea:
Re: points you brought up in your comment, but not your code review.

- I added a message letting the user know if there are now packages to be updated.

- We don't receive header information.

- The popping up buffer issue is out of scope for this issue. File a separate issue describing how we can remove that buffer when we execute a `kotct/packup` update script.

- Cleaning up packages for the users is out of scope of this issue, and the general update feature. That is something we can discuss in a separate issue

.emacs.d/lisp/package/packup.el
samontea: sure w/e

.emacs.d/lisp/package/packup.el
```diff
@@ -47,7 +229,8 @@ If UPDATE is non-nil, out-of-date packages will be updated."
                    (terpri)
                    (princ (symbol-name (car package)))
                    (princ (if (cdr package) " (update)" " (install)"))))
-               (if (y-or-n-p "Auto install/update these package?")
+               (if (or auto-update
+                       (y-or-n-p "Auto install/update these package?"))
```
rye: `Auto install/update these package(s)?`

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,199 @@
...
+    (define-key map "g" 'kotct/packup-refresh)
+    (define-key map "?" 'kotct/packup-help)
+    map))
+
+;;;###autoload
+(defun kotct/packup ()
+  "Creates an interactive buffer to install/update packages."
+  (interactive)
+  (let ((old-buf (current-buffer))
+        (buffer (get-buffer-create "*packup*")))
```
rye, line 190: screen shot 2017-08-25 at 10 31 27 am

screen shot 2017-08-25 at 10 32 06 am

I think the display problems are due to capitalization.  Can this be capitalized?

.emacs.d/lisp/package/packup.el
rye: How about adding `U` and `M`, besides?  `dired` still has a short help, and these would not be out of the scope of our short help.

rye (changes requested):
My only remaining problems are just a couple of leftovers from my previous review.

.emacs.d/lisp/package/packup.el
rye: Actually, never mind.  It's not.  It seems to be a problem with the mode.

rye: No, I revise this again.  It's because you set `mode-line-buffer-identification` to `packup` that we get rid of the bolded name.

EDIT: Added additional review for this.

.emacs.d/lisp/package/packup.el
```diff
@@ -9,17 +9,199 @@
...
+          (apply 'kotct/packup-insert-package-row (list package (package-desc-version (cadr (assq package package-alist)))))))))
+
+(defun kotct/packup-initialize-buffer ()
+  "Initializes the packup buffer."
+  (kill-all-local-variables)
+  (use-local-map packup-mode-map) ;; mapppppps
+  (setq major-mode 'packup-mode
+        mode-name "Packup"
+        buffer-read-only t ;; read only
+        mode-line-buffer-identification "packup")
```
rye: Can this be removed?  I like the default buffer ID, and it's less confusing.

rye (changes requested):

.emacs.d/lisp/package/packup.el
samontea: This is part of existing code. Open an issue.

rye (approved):
This looks good enough to me now.

cg505:
yooooooooo

PR #89, opened 2017-10-03
Add force-load to kotct/switch-to-theme

### LINKED ISSUE #88: kotct/switch-to-theme does not allow for customizations to fonts prior to load
If a theme has already been loaded, `kotct/switch-to-theme` just disables the current theme and enables the new one.

This is problematic, because then `kotct/switch-to-theme` will always just re-enable the current theme.  If, for instance, someone wants to customize `solarized`, (like turning off the variable-pitch in headings, as I prefer) they have no way of doing this.  This could be rectified by adding an optional argument to `kotct/switch-to-theme` that always loads the theme.

rye:
Adds `force-load` as an optional argument to `kotct/switch-to-theme`, which works as intended in my personal configuration.

*Closes #88.*

### DIFF
--- a/.emacs.d/lisp/visual/theme.el
+++ b/.emacs.d/lisp/visual/theme.el
@@ -7,7 +7,8 @@
 
 If FORCE-LOAD is set, THEME is always loaded."
   (if kotct/current-theme (disable-theme kotct/current-theme))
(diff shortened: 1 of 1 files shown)


rye:
Merging without review after 10+ days of inactivity.

PR #90, opened 2017-10-15
Update package archive priorities

rye:
This moves MELPA stable to the most trusted/prioritized position, and removes reference to Marmalade from the priority listing, as we have no configuration for it.

I think this might prevent collisions with versions, too.  We'll see.


rye:
Merging without approval.

PR #94, opened 2017-12-26
Darwin: force use of sh as shell for path evaluation

rye:
It is thus recommended to use a `~/.profile` file to configure your `PATH` variable, since that is what `/bin/sh` loads.  Fish shell is non-POSIX-compliant, (which is on our list of things that we cannot assume) so by using this method of gathering information, we basically are broadening the possibilities to include all Linux/Mac machines.

### DIFF
--- a/.emacs.d/lisp/system/darwin.el
+++ b/.emacs.d/lisp/system/darwin.el
@@ -45,7 +45,7 @@ stuff.  This also respects the user's shell decision."
   ;; set, and the path should already be okay.
   (if (not (getenv "TERM_PROGRAM"))
       ;; Ask the shell to print out its value for $PATH.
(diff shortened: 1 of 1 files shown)


Sammidysam (approved):

cg505:
A note on this: using fish with emacs causes other problems (such as breaking `dired-do-find-regexp` and other similar things). If your shell is fish I would recommend including in your config `(setf shell-file-name "/bin/zsh")` (or some other shell).

rye:
Yep, I still use `/usr/local/bin/fish` as my shell but I had to do a _lot_ of extra configuration stuff locally to make it happen.

cg505:
Maybe we should considering detecting if the default shell is fish and handling that by either setting the shell to zsh or adding your configuration by default

Issue #42, opened 2017-05-08
Themes are not loaded correctly in daemon mode


cg505:
If and only if emacs is started in daemon mode, the initial theme is not loaded correctly, even if the user switches to another theme and back to the first. The issues are small: colors off, etc. I noticed when using `material-theme` as used in [my config](https://github.com/cg505/.emacs).
This is probably something to do with loading the theme before we have any connection to X or something. I have not had much time to investigate the issue.

cg505:
Noticed that this issue is not specific to our config, so I opened an upstream issue at cpaulik/emacs-material-theme#45.

PR #96, opened 2018-04-29
fix themes that use (selected-frame) by manually selecting frame

cg505:
Tested and working with emacs-material-theme: see ongoing discussion at cpaulik/emacs-material-theme#45.

I don't like this fix, so I'm leaving #42 open for now; we'll see if a better fix comes along. I would like to get this merged though, so my theme works.


rye (approved):
Seems like a good idea for the time being.

PR #106, opened 2018-07-05
update repositories to https and use melpa.org

cg505:
I was having weird issues without this, not sure they are reproducible. In any case, this is a no-brainer and I would like to get it merged as quickly as possible.


rye (approved):
LGTM. :shipit:

rye:
We should be using HTTPS only by default anyways, and if the url has changed then we should use the new one.

PR #108, opened 2018-07-07
Fix packup to correctly handle melpa-stable, and other fixes

### LINKED ISSUE #99: packup does not respect `package-archive-priorities`
This needs more investigation, but it seems like our current approach in packup for checking/installing updates disregards `package-archive-priorities` (set in lisp/package/repositories.el). Namely, I see many packages on my system are installed from MELPA despite also being available in MELPA Stable (which has a higher priority).

This may be best wrapped into #84 (which would be lovely to finally get finished up when someone has a change).

cg505:
That's right: packup no longer tries to install the latest melpa even though you told it to use melpa-stable! Woo!

Other things I like are added. Feel free to complain if you don't like them.

This resolves #99.

### DIFF
--- a/.emacs.d/lisp/package/packup.el
+++ b/.emacs.d/lisp/package/packup.el
@@ -15,7 +15,7 @@
 This is what the do-commands look for, and the flag the mark-commands store.")
 
 (defmacro kotct/with-stable-package-archive-contents (body)
-  "Run BODY, but if we don't have emacs 25 or later, let-bind
+  "Run BODY, but if we don't have Emacs 25 or later, let-bind
 a modified `package-archive-contents' that only includes the package-desc
 from the most-preferred repository."
   (if (version< emacs-version "25.1")
@@ -41,6 +41,7 @@ from the most-preferred repository."
 (defun kotct/package-latest-available (package)
   "Get the lastest-available package-desc for PACKAGE, preferring repositories
 as listed in `package-archive-priorities' or `kotct/package-ordered-archives'.
+
 Does not automatically refresh the package list."
   (if (version< emacs-version "25.1")
       ;; no package-archive-priorities, so use kotct/package-ordered-archives
@@ -63,15 +64,16 @@ Does not automatically refresh the package list."
     (cadr (assoc package package-archive-contents))))
 
 (defun kotct/package-up-to-date-p (package)
-  "Returns T if PACKAGE is installed and up-to-date.
+  "Return non-nil if PACKAGE is installed and up-to-date.
+
 Does not automatically refresh package list.
-Before emacs 25, we have to manually check in preferred-repository order."
+Before Emacs 25, we have to manually check in preferred-repository order."
   (let ((latest (kotct/package-latest-available package)))
     (when latest
       (package-installed-p package (package-desc-version latest)))))
 
 (defun kotct/packup-insert-package-row (package-name package-desc-version)
-  "Inserts a package row into current buffer."
+  "Insert a package row into current buffer."
   (let ((inhibit-read-only t))
     (insert (format "[%c] %s -> %s\n" kotct/packup-marker-char package-name package-desc-version))))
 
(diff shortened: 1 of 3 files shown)


.emacs.d/lisp/package/packup.el
```diff
@@ -10,7 +10,7 @@
 
 (package-initialize)
 
-(defvar kotct/packup-marker-char ?x
+(defvar kotct/packup-marker-char ?*
```
cg505, line 13: I kept seeing `x` as the mark, so my brain would go to mark something and I'd hit `x` on my keyboard. Then it would execute the transaction. Changing to `*` avoids this issue and also brings the marks more in line with the rest of emacs (think dired, list-buffers, etc).

.emacs.d/lisp/package/packup.el
```diff
@@ -10,98 +10,149 @@
...
+    ;; package-archive-priorities means we only have to check
+    ;; the first package-desc, because that should be from the best repo
+    (cadr (assoc package package-archive-contents))))
+
 (defun kotct/package-up-to-date-p (package)
-  "Returns true if PACKAGE is up-to-date.
-Does not automatically refresh package list."
-  (every (lambda (x) (package-installed-p package (package-desc-version x)))
-         (cdr (assq package package-archive-contents))))
+  "Returns T if PACKAGE is installed and up-to-date.
```
rye: `T` &rarr; `t`?

.emacs.d/lisp/package/packup.el
```diff
@@ -10,98 +10,149 @@
...
+Does not automatically refresh package list.
+Before emacs 25, we have to manually check in preferred-repository order."
```
rye: `emacs` &rarr; `Emacs`?

rye (dismissed):
Overall LGTM, just a couple of English-related nits.

.emacs.d/lisp/package/packup.el
cg505: Actually, it should probably just be "non-nil", since that's what `package-installed-p` says in its docstring.

.emacs.d/lisp/package/packup.el
cg505: yup

cg505:
Since this PR does a LOT. I would appreciate it if you could checkout and try some things on the branch. You should be able to delete your `.emacs.d/elpa` directory entirely and get almost all packages from melpa-stable.

rye:
Okay, I did that.  I was able to install everything I needed.

~One thing is that we should make sure that `packup-update` doesn't try to update from a stable version to a non-stable version, which it seems to do.  (Try deleting `~/.emacs.d/elpa`, starting Emacs, answering `y` for everything, and then observe that `M-x kotct/packup-update` prompts you to install "newer" versions.)~

EDIT: ~I was on `master`. Boo.~ [EDIT 2:] ~I get the same behavior on `packup-fixes`, too.  I'll try nuking `.elc` files to make sure it's not that.~ [EDIT 3:] It was my `.elc` files not getting cleaned.  This problem no longer exists.

For reference, here's the [package.el](https://github.com/emacs-mirror/emacs/blob/98e98b4/lisp/emacs-lisp/package.el#L3094-L3116) source that corresponds to the <kbd>U</kbd> behavior in `package-list-packages`.

cg505:
Yeah, we should essentially end up with the same result as using `U` in the packages buffer. The only lingering issue, I think is that transitive dependencies might not have updates triggered. This is a non-trivial fix, and I'll need to spend a while figuring it out. I don't really want to hold this up with that.

So... did it work on the real branch?

rye:
Yes, I edited my comment a few times to match the current state of my hunt.  After nuking `.elc` files, I ended up getting intended results.  (<kbd>U</kbd> on `package-list-packages` reports no upgrades and `M-x kotct/packup-update` doesn't report any upgrades.)

rye (approved):
After poking around, I'm happy with using this in production.

cg505:
I am surprised nuking .elc files had an impact. We should always prefer the newer file between .el and .elc. It could be an issue with how we autoload. Probably requires further investigation in a separate issue.

I want to leave this open for at least a couple days in case @samontea has a chance to glance at it (since he wrote a lot of the code that was modified).

samontea (approved):
lgtm

cg505:
@samontea :heart_eyes:

PR #109, opened 2018-07-07
Add support for IDE-level C# editing

cg505:
I'm going to let this sit for a couple of days while I use it and try to find any remaining bugs. Feel free to review in the meantime, however.

### DIFF
--- a/.emacs.d/bin/omnisharp
+++ b/.emacs.d/bin/omnisharp
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+lsb_release -ds | grep "Arch Linux" > /dev/null
+if [ $? -ne 0 ]; then
+	echo "The omnisharp wrapper is only meant to be used with Arch Linux!"
+	echo "Please use M-x kotct/omnisharp-install-server to get omnisharp."
+	exit 1;
+fi
+
+export MONO_PATH=/opt/omnisharp-roslyn
+exec mono --assembly-loader=strict ${MONO_PATH}/OmniSharp.exe "$@"
--- a/.emacs.d/bin/omnisharp
+++ b/.emacs.d/bin/omnisharp
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-lsb_release -ds | grep "Arch Linux" > /dev/null
-if [ $? -ne 0 ]; then
+
+if ! lsb_release -ds | grep "Arch Linux" > /dev/null; then
 	echo "The omnisharp wrapper is only meant to be used with Arch Linux!"
 	echo "Please use M-x kotct/omnisharp-install-server to get omnisharp."
 	exit 1;
--- a/.emacs.d/lisp/code/languages/csharp.el
+++ b/.emacs.d/lisp/code/languages/csharp.el
@@ -1,0 +1,91 @@
+;;  -*- lexical-binding: t; -*-
+
+;;; The following keybindings are available when editing C# files using omnisharp:
+;; C-RET: Do a refactoring at point. Prompt for which refactoring in the minibuffer.
+;; M-RET: Same as C-RET
+;; C-c o r: Intelligently rename the symbol at point, throughout project
+;; M-.: Jump to definition of symbol at point
+;; C-.: Find usages of symbol at point
+;; C->: Find implementations of symbol at point
+;; C-M-\\: Reformat the region
+
+
+(defun kotct/omnisharp-install-server ()
+  "Install omnisharp-roslyn, or, if using Arch Linux, make sure it is installed
+and then install the correct wrapper."
+  (interactive)
+  (let ((os (shell-command-to-string "lsb_release -ds")))
+    (if (string= os "\"Arch Linux\"\n")
+        (if (= 0 (call-process-shell-command "pacman -Q msbuild-stable omnisharp-roslyn"))
+            (progn
+              (customize-save-variable 'omnisharp-server-executable-path "~/.emacs.d/bin/omnisharp")
+              (message "Ready to roll! Do M-x omnisharp-mode to start OmniSharp."))
+          (message "Looks like you're using Arch Linux!
+Please insall omnisharp-roslyn and msbuild-stable from the AUR.
+When you're done, try running M-x kotct/omnisharp-install-server again."))
+      ;; default to built-in installation
+      (omnisharp-install-server nil))))
+
+(defun kotct/omnisharp-maybe-start ()
+  "Start omnisharp, or prompt user to install omnisharp-roslyn."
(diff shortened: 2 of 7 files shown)


.emacs.d/bin/omnisharp
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
rye, line 1: Portability. I like it.

.emacs.d/bin/omnisharp
```diff
@@ -0,0 +1,11 @@
...
+
+lsb_release -ds | grep "Arch Linux" > /dev/null
+if [ $? -ne 0 ]; then
```
rye: Why not just merge L3 and L4 to get `if ! lsb_release -ds | grep "Arch Linux" > /dev/null;`?  I never trust `$?`, to be honest.

.emacs.d/lisp/code/languages/csharp.el
```diff
@@ -0,0 +1,91 @@
...
+and then install the correct wrapper."
+  (interactive)
+  (let ((os (shell-command-to-string "lsb_release -ds")))
+    (if (string= os "\"Arch Linux\"\n")
+        (if (= 0 (call-process-shell-command "pacman -Q msbuild-stable omnisharp-roslyn"))
+            (progn
+              (customize-save-variable 'omnisharp-server-executable-path "~/.emacs.d/bin/omnisharp")
+              (message "Ready to roll! Do M-x omnisharp-mode to start OmniSharp."))
+          (message "Looks like you're using Arch Linux!
+Please insall omnisharp-roslyn and msbuild-stable from the AUR.
```
rye: `insall` &rarr; `install`

cg505:
Also, ideally, tests (as coveralls has noted).

.emacs.d/lisp/code/languages/csharp.el
```diff
@@ -0,0 +1,91 @@
...
+;; C-.: Find usages of symbol at point
+;; C->: Find implementations of symbol at point
+;; C-M-\\: Reformat the region
+
+
+(defun kotct/omnisharp-install-server ()
+  "Install omnisharp-roslyn, or, if using Arch Linux, make sure it is installed
+and then install the correct wrapper."
+  (interactive)
+  (let ((os (shell-command-to-string "lsb_release -ds")))
```
rye, line 17: What happens if this `let` fails? E.g. `lsb_release` doesn't exist.

.emacs.d/lisp/code/languages/csharp.el
```diff
@@ -0,0 +1,91 @@
...
+  (require 'omnisharp)
+  ;; this is approximately the check omnisharp does to find the server
+  (let ((path (or omnisharp-server-executable-path
+                  (omnisharp--server-installation-path t))))
+    (if (f-exists-p path)
+        (omnisharp-mode +1)
+      (message "Heads up! For sick autocompletion and autodocs, you should install OmniSharp!
+Do M-x kotct/omnisharp-install-server and we'll take care of it."))))
+
+(add-hook 'csharp-mode-hook (lambda ()
```
rye: I forget, is our coding style to put `lambda` openings on the next line, indented to the same level as the symbol preceeding?

.emacs.d/lisp/code/smartparens-c.el
```diff
@@ -2,4 +2,13 @@
 
 (smartparens-global-mode t)
 
+(defun kotct/open-on-empty-line (&rest _ignored)
+  "Make sure the opening brace or bracket is placed on a new indented line.
+
+Intended for use as a pre-handler with smartparens."
+  (message "hi!")
```
rye: Still needed?

rye (dismissed):
LGTM with a few nits.

.emacs.d/lisp/code/languages/csharp.el
cg505: You will get an error message from the shell. Regardless, `os` won't be equal to the string indicating Arch.

.emacs.d/lisp/code/languages/csharp.el
cg505: I don't think there's a stated preference. I don't really care, so if you think it would be better that way, I'll change it.

.emacs.d/lisp/code/smartparens-c.el
cg505: fuck i knew i was going to forget this

cg505: Actually I just ended up ditching this whole function since it wasn't working as I wanted. I'm just going to delete it.

.emacs.d/lisp/code/languages/csharp.el
rye: Okay. Will this break, (i.e. throw an error and bring up an error buffer) or just complain into `*Messages*`?  (I'm not familiar enough with this to be able to know off the top of my head.)

.emacs.d/lisp/code/languages/csharp.el
rye: Yeah, we don't have a concrete coding style.  If it's not too much trouble, go ahead and try that style.

.emacs.d/lisp/code/languages/csharp.el
cg505: Nothing will happen. It will just return a different string. Just do `M-: (shell-command-to-string "foo")` in emacs right now and you'll see what it does.

rye: Gotcha.  Thanks for the clarification.

(Throwback to how I have a command called `foo` lmao)

cg505: lol, RIP my demo

rye (approved):
Nits resolved, I say this is ship ready.

cg505:
Nice. I'm going to use it at work on Monday/Tuesday and see how it's holding up, then merge.

cg505:
Can anyone jog my memory on how we are supposed to deal with tabs?

rye:
Smart-tabs by default, if I recall correctly.

cg505:
Okay, I will look into that. Also, I was looking at `kotct/setf-tab`, which I think needs a revamp. Fodder for another issue.

cg505:
We don't really need to mess with `kotct/setf-tab` in this case anyway, since csharp-mode (being a C-derived language mode) uses `c-basic-offset`.

cg505:
coming up on a year long of "a couple days" lol

rye (changes requested):
Needs merge conflicts resolved.

PR #110, opened 2018-07-07
Check unsaved buffers/active processes when killing Emacs

cg505:
I'm not sure why we weren't doing this before. It's led to lots of autosave revertions for me. I also slightly restructured the function, though the logic should be the same. Ready for review.


.emacs.d/lisp/behavior/pane-management.el
```diff
@@ -11,15 +11,12 @@ or if passed an argument, but just current frame if otherwise."
...
-            (if (not (boundp 'server-process))
-                (if (y-or-n-p "Kill Emacs?")
-                    (kill-emacs)
-                  (message "Didn't do anything, promise!"))
-              (if (y-or-n-p "Kill Emacs?")
-                  (kill-emacs)
-                (if (y-or-n-p "Delete current frame?")
-                    (delete-frame)
-                  (message "Didn't do anything, promise!"))))
+            (if (y-or-n-p "Kill Emacs?")
```
rye, line 14: Let's explain to the user that we will also save buffers before killing emacs.  (`"Save buffers and kill Emacs?"` would work fine.)

rye (changes requested):
LGTM.  I've definitely run into this issue where it just goes and kills itself and I have to load autosave files. (Which is annoying but fine.)

Granted, `save-buffers-kill-emacs` also overwrites files that are still open, which may or may not cause problems if people edit without saving and don't want to save.  (E.g. remote editing on a web server.)

We should probably express that we're going to save buffers as well, though.

cg505:
@rye `save-buffers-kill-emacs` doesn't actually save buffers without prompting. This is actually what `C-x C-c` is normally bound to in emacs. It will prompt you to save any files, then if there are still modified buffers, it will prompt. Then, if there are buffers with open processes, it will prompt. Only after you have confirmed all those prompts will it exit. It won't do anything unexpected without prompting.

rye:
My bad.

rye (approved):

PR #111, opened 2018-07-07
Let TRAMP choose its own default

### LINKED ISSUE #103: fastest TRAMP method???
we use ssh, as per emacswiki. but TRAMP itself claims scp is faster. who's right?

also TIL tramp supports adb, lol

cg505:
Per #103, don't be opinionated about the default TRAMP protocol. scp seems fine, and if you really want to, it's very easy to type `/ssh` instead of `/-`.

Closes #103.


rye (approved):
LGTM.  We'll trust the TRAMP maintainers to pick the sanest default, or our users to know how to pick their own.

PR #112, opened 2018-07-07
unsynchronize recentf saves

### LINKED ISSUE #48: Recentf Save Gets Blocked by Interlocking
In my daily life, I walk around a lot.  Emacs opens recentf a bunch, but it seems like it never really writes it to disk.  (Evidenced by the fact that `#recentf#` is always left over in `~/.emacs.d` except when the 5-minute `#'recentf-save-list` timer gets run.)

This is bad, [because Emacs has support for preventing simultaneous edits](https://www.gnu.org/software/emacs/manual/html_node/emacs/Interlocking.html), and when I change networks my hostname changes.  Emacs then blocks my workflow and asks me if I want to steal the lock from myself.  This is _highly_ frustrating.

So, should we: (a) disable interlocking support entirely, via setting `create-lockfiles` to `nil` globally, or (b) do something more intelligent?  I'm going to leave this to @cg505 because I have never used `recentf` before.

cg505:
Hopefully avoids all future recentf lock-getting issues. Closes #48.

### DIFF
--- a/.emacs.d/lisp/file/recentf-c.el
+++ b/.emacs.d/lisp/file/recentf-c.el
@@ -9,7 +9,7 @@
 (setf recentf-max-saved-items 200)
 
 ;; Set timer to save `recentf-list' every 5 minutes.
-(run-at-time t (* 5 60) #'recentf-save-list)
+(run-at-time (* 5 60) (* 5 60) #'recentf-save-list)
 
 ;; recentf completion using `ido-mode'
 ;;


cg505:
@rye if you want to give this a quick review and merge, we'll be ready to release. no rush

rye (approved):
I've also seen quoted strings instead of a number fo seconds, but this is fine.

rye:
@cg505 I will release when I get down to my coffee shop of choice

PR #93, opened 2017-12-26
Fish support

rye:
This PR adds support for editing fish configuration files.

PR #80, opened 2017-08-05
Add basic sh-mode indentation support

rye:
*Closes #69.*

~Could use some tests still.~

PR #115, opened 2018-07-15
Release v0.2.0 to stable

rye:
This release:

- Adds basic `sh-mode` indentation support. (#80)
- Includes some fixes to the functionality of the `kotct/switch-to-theme` function. (#89)
- Adds support for `yaml-mode` and `fish-mode` (#92 and #93)
- Adds support for `go-mode` (#95)
- Includes a bug fix for #48, (#112) causing `recentf` to get autosaved on a timer from the _start of the process_ rather than every _on-clock 5 minutes_, which previously caused a race condition with multiple Emacs processes.
- Includes a fix for some issues with how upgrades were selected that caused incorrect package archive priorities. (#99, fixed by #108)
- Some other bug fixes and performance improvements.

Cheers to all who participated in this release! 🎉


cg505 (approved):
:tada:

PR #118, opened 2018-07-22
Add two Emacs versions to the build matrix

rye:
I think it's one of our goals to support current versions of Emacs.

This PR simply adds Emacs versions 25.3 and 26.1 to the build matrix. Since I think we get 4x parallelism, this shouldn't slow down builds at all.

Should we also include Emacs 24 in the fun? Just to really break builds if something isn't backwards compatible.


cg505:
Take a look at the `evm` branch. I haven't opened a PR yet (because I haven't gotten around to it mostly, and packup stuff) but it is a more complete version of this PR. If you want to incorporate the additional changes from there into here, I'm cool with that too. I don't think it makes sense to test 25.1 or 25.2 because while supporting Emacs 25 is good, not many people are stuck on 25.1 and can't upgrade to 25.3. Hence also the addition of 24.5 without any early 24s.

rye:
@cg505 if it's okay with you I'll just merge `evm` into this branch and fold it all together in one PR, since those changes don't really seem out-of-scope.

I would personally like to only test three major versions, and just test the latest minor version of each of those major versions. (i.e. 24.5, 25.3, and 26.1 as of this comment.)

cg505:
Sounds like a plan.

rye:
Okay, this branch should now be good for review now.  CI is green, whatever that means.

cg505 (approved):

rye:
Merging.

PR #120, opened 2018-07-22
Add support for display-line-numbers-mode

### LINKED ISSUE #98: Transition from linum to built-in line numbers
See the help for variable `display-line-numbers` in emacs 26.

At this point we should maintain compatibility with emacs 25, but if we can use the builtin line numbers, we should, since they are faster and better.

rye:
This PR moves `linum-c` to `line-numbers-c` to make it more general-purpose, and extracts existing logic into a helper function before adding another helper method.

The new functionality on 26.1 is to disable `global-linum-mode` (probably not necessary) before enabling `global-display-line-numbers-mode`.  Then, hooks are added to turn off `display-line-numbers-mode` on a few modes that don't seem to like it, such as `dired-mode`, `git-commit-mode`, `magit`-derived modes, and `package-menu-mode`.

Open to further requests or stylistic feedback.  Otherwise, I would say ToT is ready to use in production.

Resolves #98.

### DIFF
--- a/.emacs.d/lisp/visual/line-numbers-c.el
+++ b/.emacs.d/lisp/visual/line-numbers-c.el
@@ -32,10 +32,11 @@ it generally works better."
     (global-display-line-numbers-mode 1)
 
     ;; add some hooks to turn off display-line-numbers mode
-    (add-hook 'dired-mode-hook #'kotct/display-line-numbers--turn-off)
-    (add-hook 'git-commit-mode-hook #'kotct/display-line-numbers--turn-off)
-    (add-hook 'magit-mode-hook #'kotct/display-line-numbers--turn-off)
-    (add-hook 'package-menu-mode-hook #'kotct/display-line-numbers--turn-off)))
+    (dolist (hook '(dired-mode-hook
+                    git-commit-mode-hook
+                    magit-mode-hook
+                    package-menu-mode-hook))
+      (add-hook hook #'kotct/display-line-numbers--turn-off))))
 
 (if (fboundp #'display-line-numbers-mode)
     (kotct/line-numbers--set-up-display-line-numbers)
--- a/.emacs.d/lisp/visual/line-numbers-c.el
+++ b/.emacs.d/lisp/visual/line-numbers-c.el
@@ -37,7 +37,7 @@ it generally works better."
     (add-hook 'magit-mode-hook #'kotct/display-line-numbers--turn-off)
     (add-hook 'package-menu-mode-hook #'kotct/display-line-numbers--turn-off)))
 
-(if (not (version< emacs-version "26.1"))
+(if (fboundp #'display-line-numbers-mode)
     (kotct/line-numbers--set-up-display-line-numbers)
   (kotct/line-numbers--set-up-linum))
 
--- a/.emacs.d/lisp/visual/line-numbers-c.el
+++ b/.emacs.d/lisp/visual/line-numbers-c.el
@@ -1,17 +1,24 @@
-;; enable linum globally
-(global-linum-mode 1)
+(defun kotct/line-numbers--set-up-linum ()
+  "Sets up `linum-mode' to run globally except in a few cases."
+  (progn
+    ;; enable linum globally
+    (global-linum-mode 1)
 
-;; use a lambda to generate the linum formatting
-;; this will use the minimum number of columns,
-;; but also always right-align
-(setq linum-format (lambda (line)
-                     (let* ((w (length (number-to-string (count-lines (point-min) (point-max)))))
-                            (thing (concat " %" (number-to-string (max 2 w)) "d ")))
-                       (propertize (format thing line) 'face 'linum))))
+    ;; use a lambda to generate the linum formatting
+    ;; this will use the minimum number of columns,
+    ;; but also always right-align
+    (setq linum-format (lambda (line)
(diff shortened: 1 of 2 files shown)


.emacs.d/lisp/visual/line-numbers-c.el
```diff
@@ -0,0 +1,44 @@
...
+    ;; enable linum globally
+    (global-linum-mode 1)
+
+    ;; use a lambda to generate the linum formatting
+    ;; this will use the minimum number of columns,
+    ;; but also always right-align
+    (setq linum-format (lambda (line)
+                         (let* ((w (length (number-to-string (count-lines (point-min) (point-max)))))
+                                (thing (concat " %" (number-to-string (max 2 w)) "d ")))
+                           (propertize (format thing line) 'face 'linum))))
```
cg505, line 13: not for this PR, but I think this lambda is causing a lot of performance issues in weird cases

.emacs.d/lisp/visual/line-numbers-c.el
```diff
@@ -0,0 +1,44 @@
...
+
+    ;; disable linum in certain modes
+    (require 'linum-off)
+    (add-to-list 'linum-disabled-modes-list 'package-menu-mode)
+    (add-to-list 'linum-disabled-modes-list 'magit--mode)))
+
+(defun kotct/display-line-numbers--turn-off ()
+  "Turns off `display-line-numbers-mode' for the current buffer."
+  (display-line-numbers-mode -1))
```
cg505, line 22: ... why? `(kotct/display-line-numbers--turn-off)` is literally more typing than `(display-line-numbers-mode -1)`

cg505: Oh I see, it's for the hooks. Comments there.

.emacs.d/lisp/visual/line-numbers-c.el
```diff
@@ -0,0 +1,44 @@
...
+  (progn
+    ;; inhibit global-linum-mode
+    (global-linum-mode -1)
+    (global-display-line-numbers-mode 1)
+
+    ;; add some hooks to turn off display-line-numbers mode
+    (add-hook 'dired-mode-hook #'kotct/display-line-numbers--turn-off)
+    (add-hook 'git-commit-mode-hook #'kotct/display-line-numbers--turn-off)
+    (add-hook 'magit-mode-hook #'kotct/display-line-numbers--turn-off)
+    (add-hook 'package-menu-mode-hook #'kotct/display-line-numbers--turn-off)))
```
cg505: A global function seems a little overkill for this simple thing, but idk. I might let-bind a lambda and use that instead, or dolist over the hooks and put the lambda inline (as in
```elisp
(dolist (hook '(dired-mode-hook ...))
  (add-hook hook (lambda () (display-line-numbers-mode -1))))
```
) but this seems fine. Just because it's not how I would do it doesn't make it wrong.

cg505 (dismissed):
The thing that really frustrates me about this mode is that there are spaces on both sides of the numbers in addition to the fringe. Just sitting there wasting space for no reason. Unfortunately, this is hard-coded into the Emacs C source (wtf???), so there is not much we can do there. We can, however, fix the `version<` thing. (Forgot to add this to the review, it is a single comment below.) Thanks for taking care of this.

cg505:
Forgot to hit "add" on one of the review comments, sorry. One second.

.emacs.d/lisp/visual/line-numbers-c.el
```diff
@@ -0,0 +1,44 @@
...
+
+(if (not (version< emacs-version "26.1"))
```
cg505: So I know I did this on the packup PR, but apparently it is bad style. You should check `(fboundp 'display-line-numbers-mode)` instead.

.emacs.d/lisp/visual/line-numbers-c.el
rye: I do too.

.emacs.d/lisp/visual/line-numbers-c.el
rye: Yes, that's exactly why.

.emacs.d/lisp/visual/line-numbers-c.el
rye: I'm just thinking if we end up wanting to toggle…

My big hangup on adding lambdas over functions is that I feel like this makes it easier to potentially _remove_ hooks, whereas if you set lambdas it's much harder to take care of.  Since the lambdas are doing something, and that something is something that has a pretty descriptive name, I feel like this is the best approach. (I'd obviously not do this if we had special handling such as relative line numbering for `git-commit-mode` or something ridiculous like that.)

rye: Actually, I do have an idea.  Couldn't we just do `(add-hook hook #'kotct/turn-off-line-numbers)` or whatever that is inside that enclosure?  If so, I think that's more elegant.  (More DRY.)

cg505: >this makes it easier to potentially remove hooks, whereas if you set lambdas it's much harder to take care of

This is a good point actually. I hadn't thought of this.

cg505 (approved):
sick bro

rye:
Merging.

PR #121, opened 2018-07-23
Prefer the UTF-8 coding system

rye:
Pretty self-explanatory. I can't think of a reason why not.

Emacs seems to already inherently choose UTF-8 as its default encoding, but it doesn't know that that is preferred by the user.

Do we want this, or would we rather just let Emacs make the decisions? I'm honestly not sure.

### DIFF
--- a/.emacs.d/lisp/file/file-content.el
+++ b/.emacs.d/lisp/file/file-content.el
@@ -3,4 +3,7 @@
 ;; M-x set-variable RET require-final-newline RET nil RET
 (setq-default require-final-newline t)
 
+;; UTF-8 is a sane default
+(prefer-coding-system 'utf-8)
+
 (provide 'file-content)


rye:
We could also do
```emacs-lisp
(set-language-environment 'utf-8)
(set-default-coding-systems 'utf-8)
(set-selection-coding-system 'utf-8)
(set-locale-environment "en_US.UTF-8")
```
too.

cg505:
I don't know really know much about this. I would say that if you are using this and don't experience any issues with it for at least a week, I am fine with it.

rye:
I have been using this for ~22 days and not had any problems. @cg505 / @samontea, can you review?

cg505 (approved):

rye:
Merging.

PR #122, opened 2018-09-05
web-c: Add .mjs to the list of exts for web-mode

rye:
This is the JS-person way of signifying that the file is exclusively using the "future" modules-import syntax lol

### DIFF
--- a/.emacs.d/lisp/code/languages/web-c.el
+++ b/.emacs.d/lisp/code/languages/web-c.el
@@ -2,6 +2,7 @@
 (add-to-list 'auto-mode-alist '("\\.scss\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.sass\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.css\\'" . web-mode))
+(add-to-list 'auto-mode-alist '("\\.mjs\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.jsx?\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.jsonp?\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.phtml\\'" . web-mode))


.emacs.d/lisp/code/languages/web-c.el
```diff
@@ -2,6 +2,7 @@
 (add-to-list 'auto-mode-alist '("\\.scss\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.sass\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.css\\'" . web-mode))
+(add-to-list 'auto-mode-alist '("\\.mjs\\'" . web-mode))
 (add-to-list 'auto-mode-alist '("\\.jsx?\\'" . web-mode))
```
cg505, line 6: is .mjsx a thing? if so we can combine

cg505 (approved):
lgtm

.emacs.d/lisp/code/languages/web-c.el
rye: Oh true.

Do you mean something like this?

```suggestion
 (add-to-list 'auto-mode-alist '("\\.m?jsx?\\'" . web-mode))
```

Which would obviously replace the other two… so this?

```diff
- (add-to-list 'auto-mode-alist '("\\.mjs\\'" . web-mode))
- (add-to-list 'auto-mode-alist '("\\.jsx?\\'" . web-mode))
+ (add-to-list 'auto-mode-alist '("\\.m?jsx?\\'" . web-mode))

cg505: yeah but only if .msjx is legit. if not, what you have is good

rye: It's definitely a rare thing if it is. I think we'll leave it as is for now.

rye:
Thanks for the review, @cg505!

Merging.

PR #124, opened 2018-12-24
init: Remove unnecessary call to package-initialize

### LINKED ISSUE #123: Symbol's function definition is void: package-initialize
I have gotten the following warning since switching back to Emacs' development version.

```
Warning (package): Unnecessary call to ‘package-initialize’ in init file
```

It seems that our call of this is unnecessary.

>The reason automatic package loading occurs after loading the init file is that user options only receive their customized values after loading the init file, including user options which affect the packaging system. In some circumstances, you may want to load packages explicitly in your init file (usually because some other code in your init file depends on a package). In that case, your init file should call the function `package-initialize`. It is up to you to ensure that relevant user options, such as `package-load-list`, are set up prior to the `package-initialize` call. This will automatically set `package-enable-at-startup` to `nil`, to avoid loading the packages again after processing the init file. Alternatively, you may choose to completely inhibit package loading at startup, and invoke the command <kbd>M-x package-initialize</kbd> to load your packages manually.

Given that we require any dependencies that we need to configure before configuring them, do we still need this call? Emacs is saying no.

https://github.com/kotct/dot/blob/61b9eb5833b5f31bddfc176ffead20d29c256497/.emacs.d/init.el#L6-L7

rye:
According to Emacs 26.x+, this is not necessary.  It certainly seems strange, but making this change does not affect the results of startup at all for me, besides shaving just a little bit of time off.

Since we `require` every ELPA package before we configure it, this shouldn't have any danger of affecting reachability, unless that assumption is violated anywhere.

Resolves #123.

### DIFF
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -3,9 +3,6 @@
 ;; load any absolute must-have code
 (load "~/.emacs-preload" 'noerror 'nomessage 'nosuffix)
 
-;; Initialize the package repository.
-(package-initialize)
-
 ;; Create variable to determine if we are in dev mode or not.
 ;; (Set this in your user hub.)
 (defvar kotct/dev-mode


rye:
Subsumed by #145. Closing.

PR #125, opened 2019-06-11
add interactive commands for updating kotct/dot and user configs

samontea:
Currently we only support updating user configs, and the user isn't capable of viewing the diff.

This patch allows for the user to update both user configs and kotct/dot. 
Furthermore it allows for the user to view the difference between `HEAD` and `master` before deciding whether to pull the update. 

Note: this requires a repository that doesn't require authentication to run `git fetch origin master` or `git pull origin master`.

### DIFF
--- a/.emacs.d/lisp/package/git-update.el
+++ b/.emacs.d/lisp/package/git-update.el
@@ -3,9 +3,12 @@
 ;;;  kotct/update-dot-config
 ;;;  kotct/update-user-config
 
-(require  'user-config-system)
+(require 'user-config-system)
 
 (package-initialize)
+
+(defvar kotct/directory "~/.emacs.d"
+  "The path to your instance of kotct/dot or one of it's subdirectories.")
 
 (defun kotct/git-repository-diff (directory)
   "Fetches in DIRECTORY and returns the difference between the common ancestor of HEAD and origin/master in DIRECTORY."
@@ -18,8 +21,7 @@
   (not (string= (kotct/git-repository-diff directory) "")))
 
 (defun kotct/user-fetch-config (username)
-  "Fetch USERNAME's personal config from GitHub, out of the
-repository USERNAME/.emacs."
+  "Fetch USERNAME's personal config from GitHub, out of the repository USERNAME/.emacs."
   (message "fetching config for %s" username)
   (let ((default-directory "~/.emacs.d/lisp/user/users/")
         (url (format "https://github.com/%s/.emacs.git" username)))
@@ -43,7 +45,7 @@ Returns true if an update occured."
                   (y-or-n-p (format "Pull the most recent changes to %s?" directory))))
           (if updating
               (let ((default-directory directory))
-                (kotct/run-git "pull" "origin" "master")
+                (kotct/run-git "pull" "--rebase" "origin" "master")
                 (message (format "%s updated!" directory)))
             (message "Okay we promise we didn't do anything!"))
           ;; reset buffer what not
@@ -52,6 +54,13 @@ Returns true if an update occured."
           updating))
     (progn (message (format "%s is up to date." directory))
            nil)))
+
+;;;###autoload
+(defun kotct/update-dot-config (&optional auto-update)
+  "Update USERNAME's personal config git repository if the USERNAME's personal config is out of date.
+Will auto update if AUTO-UPDATE is true."
+  (interactive "P")
+  (kotct/update-git-repository kotct/directory auto-update))
 
 ;;;###autoload
 (defun kotct/update-user-config (&optional username auto-update)
@@ -63,16 +72,8 @@ Will auto update if AUTO-UPDATE is true."
     (setf username
           (kotct/user-ask-username "Update config for: " 'require-match)))
   (if (and (kotct/update-git-repository (format "~/.emacs.d/lisp/user/users/%s" username) auto-update)
-
            (eq username kotct/user-current-username))
       ;; reload the config
       (kotct/user-switch-username username)))
-
-;;;###autoload
-(defun kotct/update-dot-config (&optional auto-update)
-  "Update USERNAME's personal config git repository if the USERNAME's personal config is out of date.
-Will auto update if AUTO-UPDATE is true."
-  (interactive "P")
-  (kotct/update-git-repository kotct/directory auto-update))
 
 (provide 'git-update)
--- a/.emacs.d/lisp/package/git-update.el
+++ b/.emacs.d/lisp/package/git-update.el
@@ -20,6 +20,13 @@
   "Fetch DIRECTORY's origin/master and diff to see if there are any new remote commits."
   (not (string= (kotct/git-repository-diff directory) "")))
 
+(defun kotct/git-current-branch (directory)
+  (let ((default-directory directory))
+    (kotct/run-git "branch" "--show-current")))
+
+(defun kotct/git-branch-is-master (directory)
+  (string= (string-trim (kotct/git-current-branch directory)) "master"))
+
 (defun kotct/user-fetch-config (username)
   "Fetch USERNAME's personal config from GitHub, out of the repository USERNAME/.emacs."
   (message "fetching config for %s" username)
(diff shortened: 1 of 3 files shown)


.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,78 @@
+;;; THIS FILE IS NOT LOADED AT STARTUP.
+;;; This file is autoloaded on the following symbols:
+;;;  kotct/update-dot-config
+;;;  kotct/update-user-config
```
rye, line 4: Nit: keep this list of autoloads in same order as `defun`s?

.emacs.d/lisp/package/package-hub.el
```diff
@@ -2,6 +2,7 @@
            (dependencies
             repositories
             verification)
-           (packup))
+           (packup
+	        git-update))
```
rye: Spaces, please.

.emacs.d/lisp/user/user-config-system.el
```diff
@@ -8,6 +8,9 @@
   "~/.emacs.d/lisp/user/default-username"
   "The file that sets the default username for the machine.  (Ignored by git.)")
 
+(defvar kotct/directory "~/.emacs.d"
```
rye: The value and description don't seem to match here—is this specifically for the `.emacs.d` or the clone of `dot`? (Does it matter?)

rye: (Also, could this also get moved to `.emacs.d/lisp/package/git-update.el` as well?)

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,78 @@
...
+(defun kotct/update-user-config (&optional username auto-update)
+  "Update USERNAME's personal config git repository if the USERNAME's personal config is out of date.
+If USERNAME is nil, prompt for a username.
+Will auto update if AUTO-UPDATE is true."
+  (interactive "P")
+  (unless username
+    (setf username
+          (kotct/user-ask-username "Update config for: " 'require-match)))
+  (if (and (kotct/update-git-repository (format "~/.emacs.d/lisp/user/users/%s" username) auto-update)
+
```
rye:
```suggestion
```

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,78 @@
+;;; THIS FILE IS NOT LOADED AT STARTUP.
+;;; This file is autoloaded on the following symbols:
+;;;  kotct/update-dot-config
+;;;  kotct/update-user-config
+
+(require  'user-config-system)
```
rye:
```suggestion
(require 'user-config-system)
```

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,78 @@
...
+          ;; erase, fill, and show new buffer
+          (erase-buffer)
+          (insert (format "%s" (kotct/git-repository-diff directory)))
+          (pop-to-buffer-same-window buffer)
+          ;; prompt user about updating and update if they want
+          (setf updating (or auto-update
+                  (y-or-n-p (format "Pull the most recent changes to %s?" directory))))
+          (if updating
+              (let ((default-directory directory))
+                (kotct/run-git "pull" "origin" "master")
```
rye: Part of me is wondering if a `pull --rebase` would be better… but this is what people would expect, to be sure.

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,78 @@
...
+    (kotct/run-git "fetch" "origin" "master")
+    (kotct/run-git "diff" "HEAD...origin/master")))
+
+(defun kotct/git-repository-is-behind (directory)
+  "Fetch DIRECTORY's origin/master and diff to see if there are any new remote commits."
+  (not (string= (kotct/git-repository-diff directory) "")))
+
+(defun kotct/user-fetch-config (username)
+  "Fetch USERNAME's personal config from GitHub, out of the
+repository USERNAME/.emacs."
```
rye: Lots of other docstrings don't wrap until the end of the sentence; is that not desired here too?

rye (commented):
Just some suggestions—I don't have much to say about the beefy logical bits here, but I definitely like where this is headed!

.emacs.d/lisp/package/git-update.el
samontea: oh. i didn't know this was a thing. i think i agree it's a better option.

done.

.emacs.d/lisp/package/git-update.el
samontea: done.

.emacs.d/lisp/package/git-update.el
samontea: done.

.emacs.d/lisp/user/user-config-system.el
samontea: uhm sure. imho we should have some higher `lisp/dot.el` in `lisp/` that contains all the "global stuff" including this and maybe `kotct/run-git` etc. this can be done in another ticket.

it's for the clone of `kotct/dot` or one of it's sub directories. we just need a directory within the git working directory. i think this is clear from the description and it's usage. 

done.

.emacs.d/lisp/package/package-hub.el
samontea: oops! done

.emacs.d/lisp/package/git-update.el
samontea: done

.emacs.d/lisp/package/git-update.el
samontea: Sorry this was just naively copied. see: https://github.com/kotct/dot/pull/125/files/443852b982f5394d004e9548c785ca2b3200612b#diff-b66ea82accaeecea8494ccaa90aec31fL39
done.

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+(package-initialize)
+
+(defvar kotct/directory "~/.emacs.d"
+  "The path to your instance of kotct/dot or one of it's subdirectories.")
+
+(defun kotct/git-repository-diff (directory)
+  "Fetches in DIRECTORY and returns the difference between the common ancestor of HEAD and origin/master in DIRECTORY."
+  (let ((default-directory directory))
+    (kotct/run-git "fetch" "origin" "master")
+    (kotct/run-git "diff" "HEAD...origin/master")))
```
rye, line 17: I am not sure if I mentioned this in my previous review, (certainly doesn't appear so) but I am not sure if this is what you want.

For reference, the docs for `git-diff(1)`:

>       git diff [<options>] <commit> <commit> [--] [<path>...]
>           This is to view the changes between two arbitrary <commit>.

>       git diff [<options>] <commit>..<commit> [--] [<path>...]
>           This is synonymous to the previous form. If <commit> on one side is
>           omitted, it will have the same effect as using HEAD instead.

>       git diff [<options>] <commit>...<commit> [--] [<path>...]
>           This form is to view the changes on the branch containing and up to
>           the second <commit>, starting at a common ancestor of both
>           <commit>. "git diff A...B" is equivalent to "git diff $(git
>           merge-base A B) B". You can omit any one of <commit>, which has the
>           same effect as using HEAD instead.

If I'm reading this right, `A...B` represents the symmetric difference between the two refs `A` and `B`, i.e. the commits present in either but not both of them. But `A..B` just works in one direction; it takes the commits in `B` that are not in `A`, so if `A` includes extra commits it still doesn't print them.

Since I think the usage of this `kotct/git-repository-diff` function (below) is just to see if there are any new changes on `origin/master`, I think the appropriate form is `..`. Otherwise, if local commits are made _ahead of_ `origin/master`, this would produce nonempty output.

rye (commented):
Looks better—one more thing from me.

cg505:
fuck i wish i had time to actually review this but thanks kris for picking up my slack. there's some chance I can take a look this weekend but don't count on it lol

.emacs.d/lisp/package/git-update.el
rye: Nope, I'm wrong. What I literally just posted says:

>`git diff A...B` is equivalent to `git diff $(git merge-base A B) B`

(`git merge-base A B` finds the oldest common ancestor.)

So yeah this is 100% right. Resolving my comment.

rye (approved):
I say ship it, but maybe give @cg505 the weekend to give it a look-over.

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+      (let ((old-buffer (current-buffer))
+            (buffer (get-buffer-create (format "*Diff %s*" directory))))
+        (set-buffer buffer)
+        (let ((inhibit-read-only t))
+          ;; erase, fill, and show new buffer
+          (erase-buffer)
+          (insert (format "%s" (kotct/git-repository-diff directory)))
+          (pop-to-buffer-same-window buffer)
+          ;; prompt user about updating and update if they want
+          (setf updating (or auto-update
```
cg505: use a `let`

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+                  (y-or-n-p (format "Pull the most recent changes to %s?" directory))))
+          (if updating
+              (let ((default-directory directory))
+                (kotct/run-git "pull" "--rebase" "origin" "master")
```
cg505: we should handle an error here (for instance, no permissions/repo DNE/no internet) and clean up appropriately

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+                (message (format "%s updated!" directory)))
+            (message "Okay we promise we didn't do anything!"))
+          ;; reset buffer what not
+          (kill-buffer buffer)
+          (set-buffer old-buffer)
+          updating))
+    (progn (message (format "%s is up to date." directory))
+           nil)))
```
cg505: you don't need to wrap this in a `progn`, since it's in the false case of the `if`

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+
+;;;###autoload
+(defun kotct/update-dot-config (&optional auto-update)
+  "Update USERNAME's personal config git repository if the USERNAME's personal config is out of date.
```
cg505: docstring is wrong

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+Will auto update if AUTO-UPDATE is true."
+  (interactive "P")
+  (kotct/update-git-repository kotct/directory auto-update))
+
+;;;###autoload
+(defun kotct/update-user-config (&optional username auto-update)
```
cg505:
```suggestion
(defun kotct/update-user-config (&optional auto-update username)
```

This way `C-u M-x kotct/update-user-config` will enable auto-update but still prompt for a username.

cg505 (changes requested):

.emacs.d/lisp/package/git-update.el
```diff
@@ -0,0 +1,79 @@
...
+  "Fetch USERNAME's personal config from GitHub, out of the repository USERNAME/.emacs."
+  (message "fetching config for %s" username)
+  (let ((default-directory "~/.emacs.d/lisp/user/users/")
+        (url (format "https://github.com/%s/.emacs.git" username)))
+    (kotct/run-git "clone" url username)))
+
+(defun kotct/update-git-repository (directory &optional auto-update)
+  "Fetches changes in DIRECTORY, and pulls from DIRECTORY if instructed by the user or AUTO-UPDATE is true.
+Returns true if an update occured."
+  (if (kotct/git-repository-is-behind directory)
```
cg505: imo we should give up right at the beginning up this function if we aren't already on master branch

.emacs.d/lisp/package/git-update.el
samontea: https://github.com/kotct/dot/blob/141796d08c3807c8a7658545c2845ff34c2f3c04/.emacs.d/lisp/user/user-config-system.el#L25-L27

This code already sends the user an error in most of those cases. I'm not sure what an appropriate cleanup operation would be (`git rebase --abort`?).

rye (approved):

.emacs.d/lisp/package/git-update.el
cg505: yeah, I guess this is probably fine. I pretty much just think that we shouldn't drop to a debugger buffer if there's no internet.

samontea: If it's invoked as an extended command through `M-x`, error will not cause a debug buffer to show. Instead it will show a message.

cg505 (approved):

cg505:
things now fail on first startup with
```
Warning (initialization): An error occurred while loading ‘~/.emacs.d/init.el’:

Symbol's function definition is void: kotct/user-fetch-config

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.
```
bc there is no autoload for kotct/user-fetch-config @samontea plz fix

`kotct/user-fetch-config` is called from `kotct/user-load-username` in user/user-config-system.el

cg505:
actually autoload is insufficient bc autoloads are loaded after all the hubs (which is where initial user setup happens) so it probably should just not be autolaod at all rip

PR #129, opened 2019-11-30
Add LSP support

rye:
This PR adds `lsp-mode` (https://github.com/emacs-lsp/lsp-mode) to the project, and demonstrates adding support for LSP in a specific language.

`lsp-mode` is a big project but seems like a _very_ solid companion for this project. In particular:

> `lsp-mode` aims to provide IDE-like experience by providing optional integration with the most popular Emacs packages like `company`, `flycheck` and `projectile`.

How do they do it? A "language server" is implemented for each language, and the "language server protocol" (JSON+RPC-based) is implemented by these servers, allowing for tasks common to all languages (code completion, hover things, jump-to-def, workspacing, and diagnostics) to be implemented once by those who know the language well.

This means that rather than write bindings for each editor for each language, (with M editors and N languages, that's M*N implementations) instead one ostensibly only needs to write one language server for each language and one client for each editor, reducing the number of implementations to M+N.

In reality, it's a bit closer to M*N+kN, where k is a small constant, as there is a bit of variability in how language servers are implemented, found, and selected between languages, but this is all just wiring effort and thus much easier to support in the long-run.

---

Some questions before I open this for merging:

- [x] Would we want to take a bold step like turning on `lsp-mode` for `prog-mode-hook` instead of just selectively enabling for certain languages? In this way, `lsp-mode` becomes the bottleneck for support, and we don't have to maintain any code. LSPs are also only used when they are found.
- [x] Do we want to be opinionated about adopting `flymake` vs `company-*`? It's been a while since I've taken a look at the differences in opinion about those. We could leave those up to the user and just use the defaults that `lsp-mode` selects, if that's an okay compromise.

### DIFF
--- a/.emacs.d/lisp/code/languages/rust.el
+++ b/.emacs.d/lisp/code/languages/rust.el
@@ -16,10 +16,4 @@
 ;; Don't indent `where' clauses in Rust.
 (setf rust-indent-where-clause nil)
 
-;; Enable LSP (to talk to RLS) on all rust-mode buffers
-;;
-;; NOTE This could get quite annoying if you're not aiming to use RLS
-(require 'lsp-mode)
-(add-hook #'rust-mode-hook #'lsp)
-
 (provide 'rust)
--- a/.emacs.d/lisp/code/languages/rust.el
+++ b/.emacs.d/lisp/code/languages/rust.el
@@ -16,4 +16,10 @@
 ;; Don't indent `where' clauses in Rust.
 (setf rust-indent-where-clause nil)
 
+;; Enable LSP (to talk to RLS) on all rust-mode buffers
+;;
+;; NOTE This could get quite annoying if you're not aiming to use RLS
+(require 'lsp-mode)
+(add-hook #'rust-mode-hook #'lsp)
+
 (provide 'rust)
--- a/.emacs.d/lisp/package/dependencies.el
+++ b/.emacs.d/lisp/package/dependencies.el
@@ -28,6 +28,8 @@
     fish-mode ;; for editing fish shell configuration files
     go-mode ;; for editing go code
     go-autocomplete ;; AC support for go
+    lsp-mode ;; for editing with any language that has a language server
+    lsp-ui ;; to support viewing lsp-bound files with flycheck and code lenses
 
     ;; THEMES
     solarized-theme)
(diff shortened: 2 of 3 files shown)


Sammidysam:
I'm not the right person to ask, getting notifications about this kinda just makes me sad

samontea:
Every time I use a new language I feel like I spend 3 hours trying to decide which language server I want to use etc. IMHO we should just leave it up to the user's config. I don't think there's a way to save a substantial amount of time with the proposed changes.

rye:
@samontea, that makes sense to me. I hadn't thought much about different language servers for the same language.

I will note that, at least on the [langserver.org](https://langserver.org/) table, it looks like a lot of languages only have one or two language server implementations to choose from, but these are things that the user might want to configure.

Would it be safe to _just_ add `lsp-mode` (also, `lsp-ui`?) to the dependencies, and the relevant line in the `.gitignore` for changes? This way, the dependencies would be installed for everyone, but not turned on.

I'm not sure about `lsp-ui` vs `company-lsp`, so maybe leaving that out and letting users decide would be better.

.emacs.d/lisp/code/languages/rust.el
```diff
@@ -16,4 +16,10 @@
 ;; Don't indent `where' clauses in Rust.
 (setf rust-indent-where-clause nil)
 
+;; Enable LSP (to talk to RLS) on all rust-mode buffers
+;;
+;; NOTE This could get quite annoying if you're not aiming to use RLS
+(require 'lsp-mode)
+(add-hook #'rust-mode-hook #'lsp)
+
```
rye: Remove these, as they are opinionated. (Not everyone would want LSP in Rust, necessarily; this should be turned on by users.)

.emacs.d/lisp/package/dependencies.el
```diff
@@ -28,6 +28,8 @@
     fish-mode ;; for editing fish shell configuration files
     go-mode ;; for editing go code
     go-autocomplete ;; AC support for go
+    lsp-mode ;; for editing with any language that has a language server
+    lsp-ui ;; to support viewing lsp-bound files with flycheck and code lenses
```
rye: Remove `lsp-ui` from this list?

rye (commented):
Self-review for the changes I'll make here.

rye:
So since this PR literally just contains the addition of `lsp-mode` as a dependency and one new line to the `.gitignore`, could we merge it?  Should we?

cg505 (approved):
go for it, sorry i haven't had any capacity to maintain this recently

PR #130, opened 2019-11-30
travis: Fix CI Errors, Upgrade Emacs

rye:
Emacs 26.3 was released on 2019-08-28, so probably a good idea to test against it now.


cg505 (approved):

rye:
I'mma just merge this and then laugh at Emacs 25.3 for a while.

We can always remove it from the build matrix later.

PR #132, opened 2020-05-29
allow changing font size while viewing read-only files

cg505:
Honestly not a fan of how we've set up font-size stuff in general, especially
since it keeps you from changing the font size for a specific buffer. But this
bug has been really pissing me off lately.


rye (approved):
LGTM!

If you want to file a follow-up issue on allowing separate font sizes per-buffer (or, just going to the default system for Emacs), I'd be happy to explore that over the next week.

PR #133, opened 2020-06-05
remove 25.3 from our travis b/c we no longer support it.

samontea:


rye (approved):
I committed this on a different branch and then got distracted 🙃.

LGTM. If evm publishes a -travis-xenial image we might want to go back to testing it, but the latest Emacs is 26 and that's really what matters most.

PR #134, opened 2020-06-06
remove indentation configuration, smart tabs customizations

samontea:
We're currently in a funny game where we're trying to set sane defaults for languages, but also respect editorconfig configurations. However, we end up failing to respect editorconfig configurations in certain circumstances.

I think we should instead try to switch to just using editorconfig to set our indentation configurations. 

If we want we can create a default .editorconfig file with sane indentation defaults and symlink it to ~/.editorconfig to give the user sane defaults for most languages.


rye (commented):
I'd re-title this PR as "Remove per-language indentation configuration, smart tabs customizations", as this PR doesn't (yet) touch the defaults specified in `indentation.el` to delete our changed defaults.  I think we should completely remove `kotct/tab-variable-setters` and all of its related machinery, relying on language-specific implementations (and `editorconfig-mode`) to keep things as they should be.  I do think the default value for `indent-tabs-mode` is fine, but other things will need to go to prevent clobbering.

It's worth noting to the skeptics here that [`editorconfig` already has a lot of support for a lot of languages](https://github.com/editorconfig/editorconfig-emacs/blob/9a73ff743b9fe820301ea9b5c5c7804bfc967324/editorconfig.el#L165-L266), and probably does a better job of keeping track of variables.  Its configuration certainly was a drop-in replacement for me with Ruby code.  We would still need language support files for the setup of additional tooling for each language, but no longer would indentation machinery matter.

I think it might also be smart to emit a warning if an EditorConfig is not found.  The sexp `(editorconfig-core-get-nearest-editorconfig default-directory)` will return the filename of the nearest `.editorconfig` file, but I think I'd also like a warning to show up if an EditorConfig exists but does not apply to the current file.  That way, we can warn the user if they're using Emacs' defaults, which I think is a scary enough thing to warrant warnings. 😄

rye:
After a bit of fiddling around, I've found a couple of helpful things that we could use to better support EditorConfig:

- `(editorconfig-get-properties)` is responsible for finding and parsing the nearest EditorConfig file's rules for the current buffer.  It returns the results in a hashtable, which is somewhat spooky but quite efficient.
- `(hash-table-count <table>)` gives us a count of the values in a hash table.
- So, if we take `(hash-table-count (editorconfig-get-properties))`, we'll get the number of properties applying to the current buffer.
  - We could then warn the user if the number of properties applying to the current buffer is `0`&mdash;then hide that message behind some `kotct/warn-user-if-editorconfig-doesnt-apply` `defvar`.

@samontea, would you mind if I mess around with this branch?

samontea:
Yeah feel free. Those changes sound good to me.

.emacs.d/lisp/code/editorconfig-c.el
```diff
@@ -27,4 +27,28 @@ Debug warning is suppressed if SUPPRESS is non-nil."
 
 (kotct/editorconfig-check-for-core kotct/editorconfig-suppress-core-not-found-warning)
 
+(defvar kotct/warn-on-editorconfig-with-no-props
+  t
+  "If not false-y, warn when editing a file with no editorconfig properties
+applying, indicating that the user is stuck with Emacs' default style for the
+active buffer.")
+
+(defun kotct/check-editorconfig-props (props)
```
samontea, line 36: i feel like this is likely testable, and therefore we should write a test for it. just create like `fakefile.fakefiletype` buffer and ensure that it warns.

samontea (commented):
your changes look good. i'd just like a little test coverage if it's trivial/doable.

.emacs.d/lisp/code/editorconfig-c.el
rye: I'm finally working on these tests; should have them ready today.

rye (commented):
I think that if we're moving the smart tabs customizations per-language, we should also remove smart tabs support as a whole (i.e. remove it from the list of `kotct` dependencies and remove our logic that interacts with it). Otherwise we're effectively just hobbling smart tabs for the languages we add support for.

(I'm fine with removing the cases where we explicitly enable `indent-tabs-mode` and instead requiring `indent-tabs-mode` to be turned on by `editorconfig`. `smart-tabs-mode` *should* be enabled when `indent-tabs-mode` is.)

.emacs.d/lisp/code/editorconfig-c.el
rye: They have been pushed with the latest commit fb8d701.

rye (approved):

cg505 (commented):
can we remove `smart-tabs-mode` from dependencies.el with this? if yes, should probably just do that and merge. i think everyone is on the same page here

PR #136, opened 2020-07-16
init: Require 'package before calling package-initialize

### LINKED ISSUE #135: config-compilation: Symbol's function definition is void: package-initialize
On the latest version on `master`, my `*config-compilation*` buffer contains the text:

```
Symbol's function definition is void: package-initialize

Process config-compilation exited abnormally with code 255
```

Do we need to `(require 'package)` before we call `(package-initialize)`?

rye:
Closes #135.

I'll have another PR to fix other things that depends on this change coming along, but this is a pretty straightforward change.

### DIFF
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -4,6 +4,7 @@
 (load "~/.emacs-preload" 'noerror 'nomessage 'nosuffix)
 
 ;; Initialize the package repository.
+(require 'package)
 (package-initialize)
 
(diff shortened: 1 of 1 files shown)


rye:
I'm going to set this as a draft for now until we resolve [the question](https://github.com/kotct/dot/issues/137#issuecomment-663091394) about loading everything at startup/bytecomp time versus loading as needed.

PR #143, opened 2020-08-20
fix: dependency issues introduced in #125

samontea:
This fixes the [dependency cycle](https://github.com/kotct/dot/pull/125#issuecomment-670779116) introduced in #125.

cc: @cg505


cg505 (approved):
idk if it work then sure

PR #145, opened 2020-08-23
fix: #144 initialize package iff <27.1

### LINKED ISSUE #123: Symbol's function definition is void: package-initialize

### LINKED ISSUE #144: Make package-initialize conditional on package being uninitialized.
This is to avoid a [new warning because of 27.1](https://github.com/jkitchin/scimax/issues/194#issuecomment-385437906).

samontea:
This is to fix a [warning introduced in emacs 27.1](https://github.com/jkitchin/scimax/issues/194#issuecomment-385437906).

EDITs by @rye:
Closes #144. 
Closes #123.

### DIFF
--- a/.emacs.d/init.el
+++ b/.emacs.d/init.el
@@ -3,8 +3,10 @@
 ;; load any absolute must-have code
 (load "~/.emacs-preload" 'noerror 'nomessage 'nosuffix)
 
+(require 'package)
 ;; Initialize the package repository.
-(package-initialize)
+(when (version< emacs-version "27.1")
+  (package-initialize))
 
 ;; Create variable to determine if we are in dev mode or not.
 ;; (Set this in your user hub.)
@@ -96,7 +98,8 @@ and we need to remove the hook and specifically use the frame."
 
 ;;; Asynchronous Byte Compilation
 (let* ((to-eval `(let ((default-directory "~/.emacs.d/lisp/"))
-                   (package-initialize)
+                   (when (version< emacs-version "27.1")
+                     (package-initialize))
                    (add-to-list 'load-path default-directory)
                    (normal-top-level-add-to-load-path ',kotct/hub-list)
                    (batch-byte-compile t)))
--- a/.emacs.d/lisp/package/git-update.el
+++ b/.emacs.d/lisp/package/git-update.el
@@ -3,8 +3,6 @@
 ;;;  kotct/update-dot-config
 ;;;  kotct/update-user-config
 
-(package-initialize)
-
 (defun kotct/run-git (&rest args)
   "Run a git command, specified by ARGS."
   (let ((process-environment (cons "GIT_TERMINAL_PROMPT=0" process-environment)))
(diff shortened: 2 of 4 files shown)


rye (commented):
Beautiful, thanks! We also have a `(package-initialize)` line in `init.el`, could you also do the same there?

rye:
I have edited the OP to include closes statements for #144 and also #123 which is basically the same issue.

It may also be worth noting that [`package-initialize` is not really necessary anymore](https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS?id=444930104e09b92ac27092aea44985637a39c26d#n172):

> Installed packages are now activated *before* loading the init file. As a result of this change, it is no longer necessary to call 'package-initialize' in your init file.

So it may be worth locking our initialization behind a version gate as well:

```elisp
(when (version< emacs-version "27.1")
  (package-initialize))
```

I'm… pretty sure that if we `package-initialize` in that way in our init file, the other `package-initialize` calls can be safely removed.

samontea:
I'd prefer to use the variable `package--intialized` because even independent of these changes [multiple `package-initialize` calls are unadvisable](https://github.com/emacs-mirror/emacs/blob/a4a3c92e9de59bd0251f36326375cce898919edc/lisp/emacs-lisp/package.el#L1460-L1465).

samontea:
haha jk. I switched to version checking. `package--intialized` didn't do what I wanted it to.

.emacs.d/init.el
```diff
@@ -3,8 +3,10 @@
 ;; load any absolute must-have code
 (load "~/.emacs-preload" 'noerror 'nomessage 'nosuffix)
 
+(require 'package)
```
rye, line 6: I'm not sure if we actually need to `(require 'package)`, but I have no issues with doing so besides potential performance losses.

rye (approved):

.emacs.d/init.el
samontea: it errored when i didn’t ¯\\_(ツ)_/¯ you can see that we’ve had to do this in [other places](https://github.com/kotct/dot/commit/accd55d87513bcbc4a158d27f55b7b456270ae5d#diff-9ff524a8cb0b974bb13c37be40048a4eL8)

PR #146, opened 2020-08-23
fix #141 remove depricated cl, replace w/ cl-lib

### LINKED ISSUE #141: Package cl is deprecated (as of Emacs 27)
With Emacs 27.1, a message is now emitted on startup:

```
Package cl is deprecated
```

We should migrate lines like `(require 'cl)` to `(require 'cl-lib)`.

samontea:
This commit removes the deprecated references to `cl` from our configuration and replaces them with new calls to function in `cl-lib`. 

This should work modulo potential differences between the old and new `cl-member` functions.

### DIFF
--- a/.emacs.d/lisp/package/packup.el
+++ b/.emacs.d/lisp/package/packup.el
@@ -6,7 +6,7 @@
 
 (require 'dependencies)
 (require 'package)
-(require 'cl)
+(require 'cl-lib)
 
 (package-initialize)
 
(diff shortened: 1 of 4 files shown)


.emacs.d/lisp/package/packup.el
```diff
@@ -27,7 +27,7 @@ from the most-preferred repository."
                                        (car (cl-member pinned-archive (cdr pkg)
```
samontea, line 27: This is the one line I'm uncertain about. [Do we believe that it still does the correct thing?](http://git.savannah.gnu.org/gitweb/?p=emacs/elpa.git;a=blob;f=packages/cl-lib/cl-lib.el;h=0f05c3e65efe8ccf26db4c8b554133b0696ba7c4;hb=HEAD#l109)

rye: The docs [here](https://www.gnu.org/software/emacs/manual/html_node/cl/Lists-as-Sets.html#Lists-as-Sets) seem to indicate that `cl-member` hasn't really changed semantically—it looks like it just finds a member (in this case `pinned-archive`) of a list (in this case `(cdr pkg)`).

rye (approved):
This looks all good to me.

We will still get `Package cl is deprecated` in instances where we require dependencies that haven't migrated, but those should (ideally) go away with time.

samontea:
cool that was my understanding. merged.

PR #149, opened 2024-01-31
change `languages` defvar to `kotct/languages`

cg505:
this breaks lsp-mode here
https://github.com/emacs-lsp/lsp-mode/blob/8105218b383c1a8a209a5f3348956a0b51446d43/lsp-mode.el#L4998
`languages` refers to the defvar instead of the local var for some reason

in retrospect, naming a global var `languages` probably wasn't the smartest


rye (approved):
>in retrospect

Is "almost a decade" old enough to be considered legacy? ;)

cg505:
oh god

PR #153, opened 2025-02-25
fix editorconfig on emacs 30

cg505:
Emacs 30 ships with a totally different version of editorconfig-mode that appears to be derived from the original package but with all non-copyright-assigned code removed. See
https://github.com/emacs-mirror/emacs/commit/f563f0f961068e5732b9770908ebfd915a93fe12

Which is kind of insane, but whatever. This PR fixes a crash on Emacs 30 that results from differences in the packages.


rye (approved):

cg505:
oh fboundp is for functions not defvars... so this patch does fix it but it will just always disable the check

